PR #16, opened 2018-09-24
Update README.md

pfvilquin:
RSA config.json botUserName was missing a capital "N" the config reader was bugging.
Also the config reader is not capable to handle if the path has a trailing "/" or not , I so added this comment:
(make sure that your "botCertPath" or "botRSAPath" include a trailing "/")

Suggested for the newbies to adjust the paths that are hardcoded in the example

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ The Python client is built in an event handler architecture. If you are building
 
 
 ## Configuration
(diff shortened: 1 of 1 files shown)


dcragusa:
I also ran into this issue.

PR #30, opened 2019-05-01
make chatbot dir into module

wcy:


karlmoll (commented):
looks good. needed for examples to work

PR #51, opened 2019-11-07
Introduce asynchronous datafeed listening service

3tilley:
The current datafeed listening service uses requests to wait on
datafeed/{id}/read until a response is received, handle that and then
wait for further events. This means DatafeedEventService.read_datafeed
is a blocking method and nothing outside of handling events can be
performed by the program. It also means that messages must be received
and handled in order, and no further messages can be processed until
the current one is done. This limitation could be handled by threads
but they are error-prone in Python, and it's not clear if the codebase
is threadsafe.

Instead this commit adds another class, AsyncDataFeedEventService that
uses aiohttp and asyncio to listen to the datafeed in an asynchronous
manner, meaning other operations can be queued on the event loop. Since
the event loop runs in on the main thread, and typically bots are
IO-bound, this allows for a model where bots can receive events,
dispatch them to other async methods and quickly return to responding
to messages. The original datafeed reading service can still be used,
but for the price of some added complexity the asychronous version
allows for far higher throughput and responsiveness.

Several other additions have been made:
  - Update the error handling to hand 400 and >=500 errors from
    the datafeed, this now matches the Java implementation
  - Add a timeout to datafeed errors
  - Add the healthcheck client
  - Improve configuration loading to be less dependent on current
    directory and clearer about entries that don't exist
  - Add a testing framework for the datafeed
  - Add missing dependencies to setup.py


3tilley:
Hello again!

From our earlier discussion there are a couple of tasks on me:
- [x] Check a potential SSL error
- [x] Fix `read_datafeed` returns a list of a list, instead of just a list
- [x] Implement asychronous versions of all the messaging functions

Then it's up to you guys to decide:
- [ ] Whether you want to release straight to 0.2.0 or do an alpha release, README needs to be updated to reflect whichever
- [x] Should we skip the side-effecting tests by default?

There were a couple of things I wondered if we could do, but didn't want to enact unilaterally:
- Should there really be logs checked in? The logging function creates a directory now if it doesn't exist, so I'd be tempted to add to `.gitignore`
- Unlike Java you can have multiple classes, so the exceptions could all live in a single file. This would be a breaking change to anyone who imports them
- In the Java codebase all the exceptions inherit from one single one, this could be done without being a breaking change
- Is it worth removing the print statements from the tests? I can't see what they're adding atm, but maybe they have another use
- Java doesn't have RSAPath in it's config, but has some other paths that Python doesn't. Is it worth unifying those?
- Not essential but do you have a view on an autoformatter? Kind of like Prettier for Javascript there is [black](https://github.com/psf/black), [yapf](https://github.com/google/yapf) (which is what Pycharm uses under the hood), [autopep8](https://pypi.org/project/autopep8/). We use black and like it, but you could pick anything.
- There may be some Python logging tricks you can employ, so that the name of every method doesn't have to be repeated in every logging line. It's also traditional to use a local logger in a method that outputs lots of logs. `log = logging.getLogger(__name__)` is the convention, and then  `log.info("foo")`. That way someone could just tune down the datafeed logs while keeping everything else at debug, for example.

3tilley:
Hello, I think this is ready for review!

3tilley:
Hello!

Following our discussion I've done some work looking into how to prevent the unpleasant logging errors. It seems the easiest way is just to manually close the session. This has to be done in a bit of a roundabout way but I believe the change below is sufficiently robust, it's passed my testing anyway. There are some other comments that I'll put individually on the PR, but this patch should fix the immediate issue. Copy the below into a text file (e.g. called `0001-Fixes-for-Reed-s-bug.txt`) and then apply with `git apply 0001-Fixes-for-Reed-s-bug.txt`.

```
From 42021a3c67d3bf26135e9671e513a0a7c20ed97b Mon Sep 17 00:00:00 2001
From: "Tilley, Max: Markets (LDN)" <Max.Tilley@barclays.com>
Date: Thu, 6 Feb 2020 17:14:19 +0000
Subject: [PATCH] Fixes for Reed's bug

---
 .../clients/sym_bot_client.py                 | 31 ++++++++++++++++---
 .../datafeed_event_service.py                 | 18 ++++++++---
 2 files changed, 41 insertions(+), 8 deletions(-)

diff --git a/sym_api_client_python/clients/sym_bot_client.py b/sym_api_client_python/clients/sym_bot_client.py
index 6f2ff0e..3b1ef9b 100644
--- a/sym_api_client_python/clients/sym_bot_client.py
+++ b/sym_api_client_python/clients/sym_bot_client.py
@@ -54,14 +54,15 @@ class SymBotClient(APIClient):
         self.health_check_client = None
 
 
-    def get_datafeed_event_service(self):
+
+    def get_datafeed_event_service(self, *args, **kwargs):
         if self.datafeed_event_service is None:
-            self.datafeed_event_service = DataFeedEventService(self)
+            self.datafeed_event_service = DataFeedEventService(self, *args, **kwargs)
         return self.datafeed_event_service
 
-    def get_async_datafeed_event_service(self):
+    def get_async_datafeed_event_service(self, *args, **kwargs):
         if self.async_datafeed_event_service is None:
-            self.async_datafeed_event_service = AsyncDataFeedEventService(self)
+            self.async_datafeed_event_service = AsyncDataFeedEventService(self, *args, **kwargs)
         return self.async_datafeed_event_service
 
     def get_datafeed_client(self):
@@ -343,3 +344,25 @@ class SymBotClient(APIClient):
         if self.health_check_client is None:
             self.health_check_client = HealthCheckClient(self)
         return self.health_check_client
+
+    async def close_async_sessions(self):
+        """Close the open aiohttp.ClientSession objects
+
+        In general this shouldn't be necessary as the interpreter should tear down
+        any Python objects, and subsequently the OS should terminate any other
+        resources (in this case sockets). However due to bugs in the Python standard
+        library certain situations trigger a race condition between asyncio and the
+        logging library that clobbers the stacktrace and raises a NameError. This
+        just ensures that doesn't happen and gives a cleaner output from the SDK.
+
+        For most usecases this method can be safely omitted.
+        """
+        logging.debug("Manually closing sessions")
+        if self.async_pod_session:
+            await self.async_pod_session.close()
+            self.async_pod_session = None
+
+        if self.async_agent_session:
+            await self.async_agent_session.close()
+            self.async_agent_session = None
+
diff --git a/sym_api_client_python/datafeed_event_service.py b/sym_api_client_python/datafeed_event_service.py
index 3c6df38..18dc223 100644
--- a/sym_api_client_python/datafeed_event_service.py
+++ b/sym_api_client_python/datafeed_event_service.py
@@ -25,7 +25,7 @@ EventTrace = namedtuple('EventTrace', 'message_id creation_time bot_received lis
 
 class DataFeedEventService:
 
-    def __init__(self, sym_bot_client, log_events=True, error_timeout_sec=None):
+    def __init__(self, sym_bot_client, log_events=True, error_timeout_sec=None, maximum_timeout_sec=None):
         self.bot_client = sym_bot_client
         self.datafeed_id = None
         self.datafeed_events = []
@@ -56,7 +56,10 @@ class DataFeedEventService:
         # Never wait less than this
         self.lower_threshold = self.baseline_timeout_sec
         # Raise a RuntimeError once this is exceeded
-        self.upper_threshold = None
+        if maximum_timeout_sec is None:
+            self.upper_threshold = config.data.get('datafeedEventsErrorMaxTimeout', 60)
+        else:
+            self.upper_threshold = maximum_timeout_sec
         # After every failure multiply the timeout by a factor
         self.timeout_multiplier = 2
 
@@ -414,8 +417,11 @@ class AsyncDataFeedEventService(DataFeedEventService):
 
         if not self.stop:
             self.stop = True
+
         await self.queue.put(None)
         await self.exception_queue.put(None)
+        await self.bot_client.close_async_sessions()
+
 
     async def read_datafeed(self):
         while not self.stop:
@@ -423,7 +429,11 @@ class AsyncDataFeedEventService(DataFeedEventService):
                 events = await self.datafeed_client.read_datafeed_async(self.datafeed_id)
 
             except Exception as exc:
-                await self.handle_datafeed_errors(exc)
+                try:
+                    await self.handle_datafeed_errors(exc)
+                except Exception as inner_exc:
+                    await self.deactivate_datafeed(wait_for_handler_completions=False)
+                    raise inner_exc from None
                 continue
 
             self.decrease_timeout()
@@ -467,7 +477,7 @@ class AsyncDataFeedEventService(DataFeedEventService):
             )
         except MaxRetryException as exc:
             log.error('AsyncDataFeedEventService - Bot has failed to authenticate more than 5 timess ')
-            raise
+            raise exc from None
 
         except (DatafeedExpiredException, APIClientErrorException, ServerErrorException) as exc:
             log.error(
-- 
2.17.1
```

reedvfeldman (commented):
PR Tested. Ready to merge.

PR #70, opened 2020-09-03
APP-2922-sym-parser-emojis-bug Fixed emojis bug

el-youness:
Done by checking the existence of the 'id' parameter in the 'data' node of the payload.

### DIFF
--- a/sym_api_client_python/processors/sym_message_parser.py
+++ b/sym_api_client_python/processors/sym_message_parser.py
@@ -60,6 +60,18 @@ class SymMessageParser:
     def get_stream_id(self, message_data):
         return message_data['stream']['streamId']
 
+    def get_tags(self, json_nodes, tag_type):
+        tags = []
+        # if json_nodes:
+        #     for k, v in json_nodes.items():
+        #         print(k, v)
+        # return tags
+        if json_nodes:
+            for k, v in json_nodes.items():
+                if 'id' in v and len(v['id']) > 0 and v['id'][0]['type'] == self.MENTION_TYPE:
+                    tags.append(v['id'][0]['value'])
+        return tags
+
     def get_mentions(self, message_data):
         mention_arr = []
         msg_xml = message_data['message']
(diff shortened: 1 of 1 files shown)


sym_api_client_python/processors/sym_message_parser.py
```diff
@@ -60,6 +60,14 @@ def get_im_name(self, message_data):
     def get_stream_id(self, message_data):
         return message_data['stream']['streamId']
 
+    def get_tags(self, json_nodes, tag_type):
```
symphony-hong: I would make this method private `def _get_tags(self, json_nodes, tag_type):` like what has been done in Java sdk: [getTags](https://github.com/SymphonyPlatformSolutions/symphony-api-client-java/blob/6fd04cd6f09ffa10d8926667cfe350e570bd6bc0/symphony-bdk-legacy/symphony-api-client-java/src/main/java/utils/SymMessageParser.java#L72)

sym_api_client_python/processors/sym_message_parser.py
```diff
@@ -60,6 +60,14 @@ def get_im_name(self, message_data):
...
+        tags = []
+        if json_nodes:
+            for k, v in json_nodes.items():
+                if 'id' in v and len(v['id']) > 0 and v['id'][0]['type'] == self.MENTION_TYPE:
```
symphony-hong: the condition should be: `v['id'][0]['type'] == tag_type`, isn't it?

symphony-hong: do you think we need to check: `'type' in v['id'][0]`? (and `value` as well)

symphony-hong (changes requested):
Some comments added

sym_api_client_python/processors/sym_message_parser.py
el-youness: done.

sym_api_client_python/processors/sym_message_parser.py
el-youness: Fixed.
For the question. I do not know the purpose of those conditions but I added them anyways to stay consistent with the Java SDK.

symphony-hong (approved):
LGTM!

symphony-elias (approved):
LGTM, maps the logic in legacy java SDK

PR #72, opened 2020-10-15
APP-3158: Update docstring of StreamClient#update_room

symphony-hong:
### Ticket
[APP-3158](https://perzoinc.atlassian.net/browse/APP-3158)

### Description
Update docstring

### Dependencies

### Checklist
- [ ] Referenced a ticket in the PR title and in the corresponding section

### DIFF
--- a/sym_api_client_python/clients/stream_client.py
+++ b/sym_api_client_python/clients/stream_client.py
@@ -68,7 +68,7 @@ class StreamClient(APIClient):
             "description": "testing update room function",
         }
 
-        update_room(stream_id, kwargs=room_obj)
+        update_room(stream_id, **room_obj)
         """
         logging.debug('StreamClient/update_room()')
         url = '/pod/v3/room/{0}/update'.format(stream_id)


symphony-youri (commented):
Please please commit messages are important 🙏 
Imagine you are looking at the history of this file and you see _Update docstring_
You have no idea what it means and no context

You might find the issue id from the branch (would be easier to have it the commit msg, also to generate changelogs) but then you have to find the issue (that's an extra step and maybe we lost the issue tracker or maybe you don't have access to it)

So you are lost without understanding why this change was made.
The commit message should be sufficient to understand why we made a change
For a bug what was not working before and explain how it was fixed

Here (or maybe I don't have enough experience with Python) it is difficult to understand the original issue

Is there a way we can unit test this change?

symphony-youri:
> Is there a way we can unit test this change?

My bad the change is just really about a comment so no unit test for that :) 
But providing a bit of context for the change still stands (for instance just explaining that invoking resp = bot_client.get_stream_client().update_room(sid, kwargs=room_obj)
 would not work and that the previous doc was misleading)

sym_api_client_python/clients/stream_client.py
```diff
@@ -68,7 +68,7 @@ def update_room(self, stream_id, **kwargs):
             "description": "testing update room function",
         }
 
-        update_room(stream_id, kwargs=room_obj)
```
thibauult, line 71: Is it the only place in the doc where `kwargs` needs to be replaced by `**`?

symphony-hong: I searched for all the client class and only found this one.

symphony-elias (dismissed):
LGTM

symphony-youri (approved):
Thanks

PR #80, opened 2020-12-07
PLAT-10139: Added missing event handlers in AsyncDatafeedEventService

symphony-elias:
### Ticket
[PLAT-10139](https://perzoinc.atlassian.net/browse/PLAT-10139)

### Description
Added missing `async` overrides in `AsyncDatafeedEventService`

### Checklist

### DIFF
--- a/sym_api_client_python/datafeed_event_service.py
+++ b/sym_api_client_python/datafeed_event_service.py
@@ -445,11 +445,23 @@ class AsyncDataFeedEventService(AbstractDatafeedEventService):
         for listener in self.room_listeners:
             await listener.on_room_member_promoted_to_owner(promoted_to_owner_data)
 
-    async def demoted_to_owner(self, payload):
-        log.debug('async demoted_to_Owner')
+    async def demoted_from_owner(self, payload):
+        log.debug('async demoted_from_owner')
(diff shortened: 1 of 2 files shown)


thibauult (approved):

sym_api_client_python/datafeed_event_service.py
```diff
@@ -445,12 +445,24 @@ async def promoted_to_owner(self, payload):
         for listener in self.room_listeners:
             await listener.on_room_member_promoted_to_owner(promoted_to_owner_data)
 
-    async def demoted_to_owner(self, payload):
-        log.debug('async demoted_to_Owner')
+    async def demoted_from_owner(self, payload):
```
symphony-youri, line 448: Are we breaking the API here?

symphony-elias: Not really. The `AsyncDatafeedEventService` is meant to be accessed from `SymBotClient`. These methods are called  by the `handle_events` method which references all the methods for each event type: https://github.com/SymphonyPlatformSolutions/symphony-api-client-python/blob/master/sym_api_client_python/services/abstract_datafeed_event_service.py#L37
Bottom line: this was not meant to be called by a developer as part of a public API.

PR #81, opened 2020-12-08
PLAT-10225: Fix datafeed loop crash for events without id

symphony-youri:
### Ticket
[PLAT-10225](https://perzoinc.atlassian.net/browse/PLAT-10225)

### Description
Fix datafeed loop crash for events without id

A datafeed event does not always have an id (for instance when external
users join a room) therefore we must handle that. The access to the id
is replaced with .get() to avoid the runtime error and if the id is
missing a random one is set for tracing purpose.

Added a unit test for that using IsolatedAsyncioTestCase, it is a bit
complex due to the async nature of the code but it does run part of the
datafeed loop.

### Dependencies
N/A
### Checklist
- [-] Docstrings added or updated
- [-] Updated the documentation in [docs folder](../docs)

### DIFF
--- a/sym_api_client_python/datafeed_event_service.py
+++ b/sym_api_client_python/datafeed_event_service.py
@@ -1,4 +1,5 @@
 import asyncio
+import uuid
 from functools import partial
 import logging
 import datetime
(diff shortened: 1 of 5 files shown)


sym_api_client_python/datafeed_event_service.py
```diff
@@ -317,6 +318,14 @@ def _process_full_trace(self, id):
                 log.error("Error while computing trace results for id: " + id)
                 log.exception(exc)
 
+    @staticmethod
```
symphony-hong, line 321: It's not needed to be a static method, right?

symphony-youri: It does not use the current so IDEA suggested me to make it static

symphony-hong: then I guess it should be 👍 sorry I often use vim for python :)

symphony-elias (dismissed):
Looks good !

symphony-elias (approved):
👍

symphony-youri:
Test PLAT-10225

PR #82, opened 2020-12-09
Use GH release page instead of release note

symphony-youri:
It will simplify the releases, no need to maintain changelog


symphony-hong (approved):
LGTM

PR #87, opened 2020-12-16
PLAT-10279 Fix async calls not using truststorePath

anthony-symphony:
### Ticket
PLAT-10279

### Description
Datafeed and other async calls that used aiohttp did not use the defined truststore in the config.json

### Dependencies
No additional dependencies

### DIFF
--- a/sym_api_client_python/clients/sym_bot_client.py
+++ b/sym_api_client_python/clients/sym_bot_client.py
@@ -3,7 +3,10 @@ import logging
 from json.decoder import JSONDecodeError
 
 import aiohttp
+import asyncio
+import sys
 import requests
+import ssl
 
 from .admin_client import AdminClient
 from .api_client import APIClient
@@ -51,6 +54,16 @@ class SymBotClient(APIClient):
         self.async_agent_session = None
         self.bot_user_info = None
         self.health_check_client = None
+        #Required for aiohttp to use truststore
+        if self.config.data[_TRUSTSTORE_PATH]:
+            logging.debug("Setting truststorePath for async calls to {}".format(config.data[_TRUSTSTORE_PATH]))
+            self.async_ssl_context = ssl.create_default_context(cafile=self.config.data[_TRUSTSTORE_PATH])
+        else: 
+            logging.debug("Setting truststore for async calls to system truststore")
+            self.async_ssl_context = ssl.create_default_context()
+        #Required to work around a bug in Python 3.8+ on Windows. Can be removed once fixed in aiohttp
+        if sys.version_info[0] == 3 and sys.version_info[1] >= 8 and sys.platform.startswith('win'):
+            asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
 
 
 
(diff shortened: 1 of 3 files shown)


sym_api_client_python/clients/sym_bot_client.py
```diff
@@ -51,6 +54,16 @@ def __init__(self, auth, config):
         self.async_agent_session = None
         self.bot_user_info = None
         self.health_check_client = None
+        #Required for aiohttp to use truststore
```
symphony-youri: the downside here is that we are creating the ssl context even if we are not using the async calls

but at the same time, code is factored in one place and we avoid using the session object to pass the ssl context around

sym_api_client_python/clients/sym_bot_client.py
```diff
@@ -51,6 +54,16 @@ def __init__(self, auth, config):
...
+        if self.config.data[_TRUSTSTORE_PATH]:
+            logging.debug("Setting truststorePath for async calls to {}".format(config.data[_TRUSTSTORE_PATH]))
+            self.async_ssl_context = ssl.create_default_context(cafile=self.config.data[_TRUSTSTORE_PATH])
+        else: 
+            logging.debug("Setting truststore for async calls to system truststore")
```
symphony-youri: since it will the default maybe we don't need the debug log

sym_api_client_python/clients/sym_bot_client.py
```diff
@@ -51,6 +54,16 @@ def __init__(self, auth, config):
...
+            self.async_ssl_context = ssl.create_default_context()
+        #Required to work around a bug in Python 3.8+ on Windows. Can be removed once fixed in aiohttp
+        if sys.version_info[0] == 3 and sys.version_info[1] >= 8 and sys.platform.startswith('win'):
```
symphony-youri: is there a risk that we are changing the loop policy too late? like if asyncio has already been initialized before?

I have mixed feelings on adding this in the BDK, maybe it would be safer to use it as a workaround (in the bot) when you are facing the issue?
I'm thinking for instance what if in some context setting this loop policy is the wrong thing to do?

symphony-youri (commented):
Thanks for the PR @anthony-symphony!

We will try to reproduce it on our end too. Maybe it is worth checking if the latest version of aiohttp still has the windows issue

sym_api_client_python/clients/sym_bot_client.py
symphony-elias: Maybe we could extract that to a method `get_async_ssl_context` called only where it is needed. It would create the async SSL context the first time and reuse it next times,

sym_api_client_python/clients/sym_bot_client.py
anthony-symphony: I am open to taking that out and putting it in the README.md

sym_api_client_python/clients/sym_bot_client.py
anthony-symphony: Will remove.

sym_api_client_python/clients/sym_bot_client.py
symphony-youri: @anthony-symphony Yes please I would be more comfortable with this specific workaround out of the code base
Do our customers use Windows for production or only for development?

symphony-youri (changes requested):
If you could just take the Windows workaround out to the readme please

symphony-youri (approved):

PR #88, opened 2021-01-04
Handle asyncio.CancelledError

anthony-symphony:
### Ticket
PLAT-10280

### Description
We currently do not handle asyncio.CancelledError. In certain situations this could cause the datafeed to keep reading. This fixes it by catching the CancelledError exception, stopping the datafeed and rethrowing it without recreating the datafeed.

### DIFF
--- a/sym_api_client_python/datafeed_event_service.py
+++ b/sym_api_client_python/datafeed_event_service.py
@@ -234,7 +234,10 @@ class AsyncDataFeedEventService(AbstractDatafeedEventService):
         while not self.stop:
             try:
                 events = await self.datafeed_client.read_datafeed_async(self.datafeed_id)
-
+            except CancelledError as exc:
+                log.info("Cancel request received. Stopping datafeed...")
+                await self.deactivate_datafeed()
+                raise
             except Exception as exc:
                 try:
                     await self.handle_datafeed_errors(exc)
@@ -272,10 +275,6 @@ class AsyncDataFeedEventService(AbstractDatafeedEventService):
         """
         try:
             raise thrown_exception
-        except CancelledError as exc:
-            log.info("Cancel request received. Stopping datafeed...")
-            await self.deactivate_datafeed()
-            raise
         except UnauthorizedException:
             log.error('AsyncDataFeedEventService - caught unauthorized exception')
         except MaxRetryException as exc:
--- a/sym_api_client_python/datafeed_event_service.py
+++ b/sym_api_client_python/datafeed_event_service.py
@@ -5,6 +5,8 @@ import logging
 import datetime
 from collections import namedtuple
 
+from asyncio import CancelledError
+
 #TODO: These imports are duplicated over the abstract class to avoid errors in the Async version
 from .exceptions.UnauthorizedException import UnauthorizedException
 from .exceptions.APIClientErrorException import APIClientErrorException
@@ -270,6 +272,10 @@ class AsyncDataFeedEventService(AbstractDatafeedEventService):
         """
         try:
             raise thrown_exception
+        except CancelledError as exc:
+            log.info("Cancel request received. Stopping datafeed...")
+            await self.deactivate_datafeed()
+            raise
         except UnauthorizedException:
             log.error('AsyncDataFeedEventService - caught unauthorized exception')
         except MaxRetryException as exc:


symphony-elias:
Shouldn't we catch a `KeyboardInterrupt` instead, like stated here: https://stackoverflow.com/questions/48562893/how-to-gracefully-terminate-an-asyncio-script-with-ctrl-c or https://stackoverflow.com/questions/37417595/graceful-shutdown-of-asyncio-coroutines ?

anthony-symphony:
asyncio does not throw KeyboardInterrupt. I just tested and it does not work. When I do ctrl+c nothing happens. Ctrl+C again ends the program without running the logic inside the except. For example in my program that utilizes the api-client:

    try:
        await datafeed
    except KeyboardInterrupt as ex:
        print("Exception Keyboard Interrupt")
        logging.exception("Cancelled Exception" + str(ex))
    print("after cleanup")

neither is printed nor is it logged. When I use CancelledError it works.

I was never able to reproduce the issue where the program will not stop and the exception is swallowed completely but, this is to fix the fact that we did not cleanly end the datafeed which should fix the exception getting swallowed up as well (hopefully)

sym_api_client_python/datafeed_event_service.py
```diff
@@ -270,6 +272,10 @@ async def handle_datafeed_errors(self, thrown_exception):
         """
         try:
             raise thrown_exception
+        except CancelledError as exc:
+            log.info("Cancel request received. Stopping datafeed...")
+            await self.deactivate_datafeed()
```
symphony-youri: Do we want for the pending query to complete here? Or could it be faster to pass wait_for_handler_completions=False?

symphony-youri:
Here https://stackoverflow.com/questions/30765606/whats-the-correct-way-to-clean-up-after-an-interrupted-event-loop it seems to suggest that the behavior might be different with different versions of Python. Which version are you using @anthony-symphony?

anthony-symphony:
The behavior is the same it is the way we start the loop that is different. Based on the link where they are putting except KeyboardInterrupt would be similar to what is implemented here:

https://github.com/python/cpython/blob/3.8/Lib/asyncio/runners.py

Where it catches the KeyboardInterrupt and then cancels the task. But, when it cancels the task our program will still do nothing because it throws the CancelledError but, we don't catch it until the catch all where we log it and recreate the datafeed. We can verify this with the old code as we see it hit this except block.

https://github.com/SymphonyPlatformSolutions/symphony-api-client-python/blob/628cbf3aa857d3bc66a98a3ec009ce1c82295d05/sym_api_client_python/datafeed_event_service.py#L282

When I try to catch KeyboardInterrupt instead of CancelledError I end up hitting the same block as well.

This change will catch the CancelledError, stop the datafeed and reraise it so the developer can implement their cleanup.

symphony-youness:
I ran the async loop and verified that I can send a message and receive one from the bot before hitting (Ctrl+c).
It interrupts and stops the bot with these python versions: 3.9.0 | 3.8.0 | 3.7.9 | 3.6.12
Can you give more information about your environment so that I can reproduce the issue?

Here is an example of it stopping using ctrl+c:

symphony-youness (dismissed):
Apparently they used a different way of running an async loop (which is not how we describe it in the python SDK's documentation) but it is one way to do it.
It explains the error we can't reproduce if we follow the SDK's documentation.

using: `asyncio.run(loop)`

I couldn't reproduce the hanging shell but there is indeed a non-handled `asyncio.CancelledError`

**There is no clear solution and it depends on how the keyboard interrupt is triggered:** https://github.com/ethereum/py-evm/issues/385#issuecomment-367611201

The PR does the handling of the exception it and is non-blocking for the documented way of running the async loop. 

- Here is the base issue raised on python git: https://github.com/python/asyncio/issues/341

symphony-youri:
So using this main

```
import asyncio
import logging
import os
import sys
from asyncio import CancelledError

from sym_api_client_python.auth.rsa_auth import SymBotRSAAuth
from sym_api_client_python.clients.sym_bot_client import SymBotClient
from sym_api_client_python.configure.configure import SymConfig


def configure_logging():
    logging.basicConfig(
        level=logging.DEBUG
    )
    logging.getLogger("urllib3").setLevel(logging.WARNING)


async def main():
    print('Python Client runs using RSA authentication')
    if sys.version_info[0] == 3 and sys.version_info[1] >= 8 and sys.platform.startswith('win'):
        asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
    configure = SymConfig(os.path.join(os.path.dirname(__file__), "..", "sym_api_client_python", "resources", "rsa_config.json"))
    configure.load_config()
    auth = SymBotRSAAuth(configure)
    auth.authenticate()
    bot_client = SymBotClient(auth, configure)
    datafeed_event_service = bot_client.get_async_datafeed_event_service()
    print('Starting datafeed')
    datafeed = datafeed_event_service.start_datafeed()
    print('Started datafeed')
    try:
        await datafeed
    except KeyboardInterrupt as ex:
        print("Exception Keyboard Interrupt")
        logging.exception("KeyboardInterrupt" + str(ex))
    except CancelledError as ex:
        print("Exception CancelledError")
        logging.exception("Cancelled Exception" + str(ex))
    print("after cleanup")

def main2():
    print('Python Client runs using RSA authentication')
    if sys.version_info[0] == 3 and sys.version_info[1] >= 8 and sys.platform.startswith('win'):
        asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
    configure = SymConfig('./config.json')
    configure.load_config()
    auth = SymBotRSAAuth(configure)
    auth.authenticate()

    bot_client = SymBotClient(auth, configure)
    print('Starting datafeed')
    datafeed_event_service = bot_client.get_async_datafeed_event_service()

    loop = asyncio.get_event_loop()
    awaitables = datafeed_event_service.start_datafeed()
    loop.run_until_complete(awaitables)
    print('Started datafeed')

if __name__ == "__main__":
    configure_logging()
    #main2()
    try:
        asyncio.run(main())
    except KeyboardInterrupt:
        print("keyboardinterrupt")
    print("ending")
```

and Docker / Python 3.8, I'm seeing:

```
root@1876ee2dda5a:/opt# python examples/main_async.py
DEBUG:asyncio:Using selector: EpollSelector
Python Client runs using RSA authentication
DEBUG:root:Loading config from: /opt/sym_api_client_python/resources/rsa_config.json
DEBUG:root:RSA Auth/authenticate()
DEBUG:root:RSA Auth/authenticate() --> needed to authenticate
DEBUG:root:RSA_auth/get_session_token() function started
DEBUG:root:RSA_auth/getJWT() function started
DEBUG:root:RSA/session token success
DEBUG:root:RSA_auth/get_keyauth()
DEBUG:root:RSA_auth/getJWT() function started
DEBUG:root:RSA/key manager token success
Starting datafeed
Started datafeed
DEBUG:sym_api_client_python.datafeed_event_service:AsyncDataFeedEventService/start_datafeed()---
DEBUG:sym_api_client_python.services.datafeed_id_repository:Retrieving datafeed id from file /opt/datafeed.id
DEBUG:sym_api_client_python.services.datafeed_id_repository:Retrieved datafeed id: 6af01278-b98a-4b5f-9684-f7972af4697d
DEBUG:root:DataFeedClientV1/read_datafeed_async()
DEBUG:root:bot_client/get_agent_session() - creating async agent session
DEBUG:sym_api_client_python.datafeed_event_service:Waiting on new exception, queue size: 0
^CException CancelledError
ERROR:root:Cancelled Exception
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/usr/local/lib/python3.8/asyncio/base_events.py", line 603, in run_until_complete
    self.run_forever()
  File "/usr/local/lib/python3.8/asyncio/base_events.py", line 570, in run_forever
    self._run_once()
  File "/usr/local/lib/python3.8/asyncio/base_events.py", line 1823, in _run_once
    event_list = self._selector.select(timeout)
  File "/usr/local/lib/python3.8/selectors.py", line 468, in select
    fd_event_list = self._selector.poll(timeout, max_ev)
KeyboardInterrupt

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "examples/main_async.py", line 42, in main
    await datafeed
  File "/opt/sym_api_client_python/datafeed_event_service.py", line 210, in start_datafeed
    await asyncio.gather(self.read_datafeed(), self.handle_events(), self.handle_exceptions())
asyncio.exceptions.CancelledError
after cleanup
keyboardinterrupt
ending
ERROR:asyncio:Unclosed client session
client_session: <aiohttp.client.ClientSession object at 0x7f9f72980940>
```

Should not we have a `Cancel request received. Stopping datafeed` message?

symphony-youri:
Follow-up on discussions with @anthony-symphony and @symphony-youness

We noticed a difference between 3.7 and 3.8+ but @anthony-symphony latest change fixes it.
Using Docker I was able to test it with different versions.

@symphony-elias @symphony-youness Ok for you to merge?

(@symphony-elias we should add an .circleci config it is running a build now that we configured it :))

symphony-youri (approved):

symphony-elias (approved):
Looks good !

PR #90, opened 2021-01-05
Set up PR builder

symphony-elias:
### Ticket
PLAT-10243

### Description
Set up circle CI build

### Checklist


symphony-youri (commented):
Can we enable it to see how it runs in CircleCI?

.circleci/config.yaml
```diff
@@ -0,0 +1,40 @@
...
+            - /home/circleci/.cache/pypoetry/virtualenvs
+
+      - run:
+          name: Run pylint
+          command: poetry run pylint symphony_api_client_python tests > pylint_report.txt
+
+      - run:
+          name: Run tests with coverage
+          command: |
+            mkdir test-results
```
symphony-youri: is it needed? (could it be part of the build? to avoid having to do it locally or maybe locally we would just run without args and use the console?)

.circleci/config.yaml
```diff
@@ -0,0 +1,40 @@
...
+          command: poetry install
+
+      - save_cache:
+          key: deps-{{ checksum "poetry.lock" }}
+          paths:
+            - /home/circleci/.cache/pypoetry/virtualenvs
+
+      - run:
+          name: Run pylint
+          command: poetry run pylint symphony_api_client_python tests > pylint_report.txt
```
symphony-youri: isn't the command line output enough?

symphony-youri: if we can make it work with just `poetry run pylint` that would be cool

thibauult (dismissed):
LGTM 👍

symphony-youri:
The tests do not seem to be reported in the test tab (https://app.circleci.com/pipelines/github/SymphonyPlatformSolutions/symphony-api-client-python/6/workflows/befbdaba-8ca0-4a88-a4ec-a0096b0e6f36/jobs/8/tests)

.circleci/config.yml
```diff
@@ -0,0 +1,38 @@
...
+          command: poetry run pylint symphony_api_client_python tests
+
+      - run:
+          name: Run tests with coverage
+          command: poetry run pytest
+
+      - store_test_results:
+          path: junit.xml
+
+      - store_artifacts:
```
symphony-youri, line 34: isn't store_test_results enough to view the test results?

symphony-elias: I followed this: https://circleci.com/docs/2.0/collect-test-data/#pytest
And test is not reported under the test tabs.

symphony-youri: It looks like path is a folder and not directly the junit.xml file

symphony-youness: I think it should be a folder, it says `path: test-results` but not a file.

symphony-youness: `Note: Please write your tests to subdirectories of your store_test_results path, ideally named to match the names of your particular test suites` - [source](https://circleci.com/docs/2.0/configuration-reference/#store_test_results)

symphony-youri (dismissed):
Should we add Snyk too?

symphony-youness (dismissed):
The builld is now showing tests on the test tab. I suppose the results are simple because we still don't have tests.

LGTM

.circleci/config.yml
```diff
@@ -0,0 +1,38 @@
...
+          command: poetry install
+
+      - save_cache:
+          key: deps-{{ checksum "poetry.lock" }}
+          paths:
+            - /home/circleci/.cache/pypoetry/virtualenvs
+
+      - run:
+          name: Run pylint
+          command: poetry run pylint symphony_api_client_python tests
```
symphony-hong: Regarding the idea renaming the package to `symphony.bdk`, should we consider to rename it within this PR so we won't need to update the circleCI jobs in future?

symphony-youness (dismissed):

symphony-hong (dismissed):
LGTM

symphony-youri (approved):

PR #95, opened 2021-01-12
PLAT-10245: Support CircleCI build for master branch

symphony-youri:
### Ticket
PLAT-10245

### Description

In the 2.0 branch we introduced CircleCI and PR builders, just to make
it work with the master branch, here we add a simple build file just
running the tests (the bare minimum, the 2.0 branch supports Snyk, test
coverage and reports).

3.6 cannot be used because we use unittest.async_case.
Caching with pyenv was not working (dependencies problem) so it is not
implemented (we do not expect a lot of executions).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,16 @@
+version: 2.1
+
+jobs:
+  build:
+    docker:
+      - image: cimg/python:3.9
+    steps:
+      - checkout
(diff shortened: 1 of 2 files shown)


symphony-hong (approved):
LGTM

symphony-elias (approved):
Looks good !

PR #98, opened 2021-01-15
PLAT-10366: Added generated API

symphony-elias:
### Ticket
PLAT-10366

### Description
Added generated API from swagger specs, namely:
- https://github.com/symphonyoss/symphony-api-spec/blob/master/agent/agent-api-public.yaml
- https://github.com/symphonyoss/symphony-api-spec/blob/master/authenticator/authenticator-api-public.yaml
- https://github.com/symphonyoss/symphony-api-spec/blob/master/login/login-api-public.yaml
- https://github.com/symphonyoss/symphony-api-spec/blob/master/pod/pod-api-public-deprecated.yaml

Generate api and models were put in separate python packages because of clashing tags across yaml files making some python modules to be overwritten.

Updated circle ci file to trigger snyk scan on 2.0 branch

### Checklist
- [NA] Unit tests updated or added
- [NA] Docstrings added or updated


symphony-youness (approved):
LGTM

symphony-hong:
I have a small question is that do we need to commit also the template since we decided to commit all the generated code? it's somehow redundant, isn't it?

symphony-elias:
> I have a small question is that do we need to commit also the template since we decided to commit all the generated code? it's somehow redundant, isn't it?

Yes but this is just in case we want to regenerate then files when swagger specs change. One never knows and I prefer to keep it.

symphony-hong (approved):
LGTM

thibauult (approved):

symphony-youri (approved):

PR #99, opened 2021-01-18
PLAT-10242 Configuration

symphony-youness:
### PLAT-10242
[PLAT-10242](https://perzoinc.atlassian.net/browse/PLAT-10242)

### Description
Added the configuration model following the model implemented in Java BDK2.0: 
[Java BDK model](https://github.com/SymphonyPlatformSolutions/symphony-api-client-java/tree/master/symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model)

I didn't implement the parameter interpolation in the BdkConfigParser

The BdkConfig class contains:
- [x] agent
- [x] pod
- [x] key_manager
- [x] session_auth
- [x] bot
- [x] ssl
- [ ] datafeed
- [ ] app
- [ ] retry
- [ ] proxy

### Checklist


symphony/bdk/core/config/bdk_config_loader.py
```diff
@@ -0,0 +1,49 @@
...
+class BdkConfigLoader:
+
+    @classmethod
+    def load_from_file(cls, config_path: str):
+        config_path = Path(config_path)
+        if config_path.exists():
+            config_content = config_path.read_text()
+            return cls.load_from_content(config_content)
+        else:
+            raise BdkConfigException("Config file has not been found at: %s" % config_path.absolute())
```
symphony-elias: Please use f-strings: https://realpython.com/python-string-formatting/

symphony/bdk/core/config/bdk_config_loader.py
```diff
@@ -0,0 +1,49 @@
...
+        if data_dict is not None:
+            return BdkConfig(**data_dict)
+        else:
+            raise BdkConfigException("Config file exists but is empty")
+
+    @classmethod
+    def load_from_symphony_dir(cls, relative_path: str):
+        """Load BdkConfig from a relative path located in the .symphony directory.
+
+        Note: The .symphony directory is located under your home directory (``System.getProperty("user.home")``).
```
symphony-elias: `System.getProperty("user.home")` is java-specific, IMHO you can remove that as `home directory` is clear enough

symphony/bdk/core/config/bdk_config_loader.py
```diff
@@ -0,0 +1,49 @@
...
+        It's a convention adopted in order to avoid storing sensitive information (such as usernames, private keys...)
+        within the code base.
+
+        :param relative_path: configuration's relative path from the ``${user.home}/.symphony`` directory
+        :return: Symphony bot configuration object
+        """
+        home_path = Path.home()
+        symphony_dir_path = home_path / ".symphony"
+        config_path = (symphony_dir_path / relative_path).resolve()
```
symphony-elias: IMHO, you could inline home_path and symphony_dir_path and put directly: `config_path = (Path.home() / ".symphony" / relative_path).resolve()

symphony/bdk/core/config/bdk_config_loader.py
```diff
@@ -0,0 +1,49 @@
...
+
+        if config_path.exists():
```
symphony-elias: Please reuse `load_from_file`

symphony/bdk/core/config/bdk_config_parser.py
```diff
@@ -0,0 +1,25 @@
...
+            return json.loads(config_content)
+        except json.JSONDecodeError:
+            pass
+
+        try:
+            return yaml.safe_load(config_content)
+        except yaml.YAMLError as e:
+            pass
+
+        raise BdkConfigException("Config file is not the right file type")
```
symphony-elias: Please be more specific with something like : `Config file format is neither JSON nor YAML"

symphony/bdk/core/config/model/bdk_agent_config.py
```diff
@@ -0,0 +1,7 @@
+from symphony.bdk.core.config.model.bdk_client_config import BdkClientConfig
```
symphony-elias: If I remember well, this class was used to accommodate for the load-balancing usecase. So you can remove it,

symphony/bdk/core/config/bdk_config_loader.py
```diff
@@ -0,0 +1,49 @@
...
+            return cls.load_from_content(config_content)
+        else:
+            raise BdkConfigException("Config file has not been found at: %s" % config_path.absolute())
+
+    @classmethod
+    def load_from_content(cls, content: str) -> BdkConfig:
+        return cls.parse_config(BdkConfigParser.parse(content))
+
+    @staticmethod
+    def parse_config(data_dict: dict):
```
symphony-elias: IMHO this belongs to the BdkConfigParser class.

symphony/bdk/core/config/bdk_config_parser.py
symphony-youness: I will change it to `Config file has a wrong format.` as it can be a misformatted YAML file. 
There is a test using an invalid yaml file btw.

symphony/bdk/core/config/model/bdk_agent_config.py
symphony-youness: I shouldn't have pushed this in the first place... 😄

symphony/bdk/core/config/bdk_config_loader.py
symphony-youness: This was done explicity, to give another error message when loading from symphony_dir:
"Unable to load the configuration file from .symphony directory." 
vs a normal

symphony/bdk/core/config/bdk_config_parser.py
symphony-youness: We can add that information in the logs. After implementing that.

symphony/bdk/core/config/model/bdk_authentication_config.py
```diff
@@ -0,0 +1,31 @@
...
+    def __init__(self, private_key=None, certificate=None):
+        self._bdk_rsa_key_config = BdkRsaKeyConfig(**private_key) if private_key is not None else BdkRsaKeyConfig()
+        self._bdk_certificate_config = BdkCertificateConfig(**certificate) if certificate is not None else BdkCertificateConfig()
+
+    def is_rsa_authentication_configured(self) -> bool:
+        """Check if the RSA authentication is configured
+
+        :return true if the RSA authentication is configured
+        """
+        return (self.bot_rsa_key_config is not None and self.bot_rsa_key_config.is_configured())
```
symphony-elias: `self.bot_rsa_key_config is not None` will always be true because of initialization in constructor.

symphony/bdk/core/config/model/bdk_bot_config.py
```diff
@@ -0,0 +1,21 @@
+from symphony.bdk.core.config.model.bdk_authentication_config import BdkAuthenticationConfig
+
+
+class BdkBotConfig(BdkAuthenticationConfig):
+
+    def __init__(self, config):
+        if config is not None:
+            self.__username = config.get("username")
```
symphony-elias: Please use a single underscore for the username field.

symphony/bdk/core/config/model/bdk_certificate_config.py
```diff
@@ -0,0 +1,46 @@
+class BdkCertificateConfig:
+
+    def __init__(self, path=None, content="", password=None, **kwargs):
```
symphony-elias, line 3: `**kwargs` is useless here.

symphony/bdk/core/config/model/bdk_certificate_config.py
```diff
@@ -0,0 +1,46 @@
...
+        self._password = password
+
+    def is_configured(self) -> bool:
+        """"Check if the certificate authentication is configured or not
+
+        :return: true if the certificate authentication is configured
+        """
+        return self.path is not None or self.content
+
+    def is_valid(self) -> bool:
```
symphony-elias, line 16: Do we need two functions for the same check?

symphony/bdk/core/config/model/bdk_client_config.py
```diff
@@ -0,0 +1,42 @@
+from symphony.bdk.core.config.model.bdk_server_config import BdkServerConfig
+
+
+class BdkClientConfig(BdkServerConfig):
+
+    def __init__(self, parent_config, config):
+        self.parent_config = parent_config
+        if config is not None:
+            self._scheme = config.get("scheme") if "scheme" in config else None
```
symphony-elias: https://docs.python.org/3/library/stdtypes.html#dict.get : get("key") on a dict defaults to `None` id key is not present.

symphony/bdk/core/config/model/bdk_client_config.py
```diff
@@ -0,0 +1,42 @@
...
+
+    @property
+    def context(self):
+        return self.__self_or_parent(self._context, self.parent_config.context)
+
+    @property
+    def proxy(self):
+        return self.__self_or_parent(self._proxy, self.parent_config.proxy)
+
+    def __self_or_parent(self, instance_value, parent_value):
```
symphony-elias: Please use one single underscore for private methods. Please check for the difference between single underscore and double underscores: https://stackoverflow.com/questions/1301346/what-is-the-meaning-of-single-and-double-underscore-before-an-object-name

symphony/bdk/core/config/model/bdk_config.py
```diff
@@ -0,0 +1,55 @@
+from symphony.bdk.core.config.model.bdk_bot_config import BdkBotConfig
+from symphony.bdk.core.config.model.bdk_server_config import BdkServerConfig
+from symphony.bdk.core.config.model.bdk_client_config import BdkClientConfig
+from symphony.bdk.core.config.model.bdk_ssl_config import BdkSslConfig
+
+
+class BdkConfig(BdkServerConfig):
+    """Class containing the Bdk configuration """
+
+    def __init__(self, **kwargs):
```
symphony-elias: Please put directly constructor params `scheme`, `host`, `port`, etc. before **kwargs and remove `kwargs.get`
The code below works:
```python
def _private(a, b, c, **kwargs):
    print(f'a={a}, b={b}, c={c}')
    print(kwargs)

def test_unpack():
    d = {'a': "value a", 'b': "value b", 'c': "value c", 'd': 'toto'}
    _private(**d)
```

symphony/bdk/core/config/model/bdk_config.py
```diff
@@ -0,0 +1,55 @@
...
+        self.__agent = BdkClientConfig(self, kwargs.get("agent"))
+        self.__pod = BdkClientConfig(self, kwargs.get("pod"))
+        self.__key_manager = BdkClientConfig(self, kwargs.get("keyManager"))
+        self.__session_auth = BdkClientConfig(self, kwargs.get("sessionAuth"))
+
+        self.__bot = BdkBotConfig(kwargs.get("bot"))
+        self.__ssl = BdkSslConfig(kwargs.get("ssl"))
+
+    @property
+    def agent(self): return self.__agent
```
symphony-elias: If we don't need any specific logic in getters and setters, IMHO we can use fields without underscores and remove the properties

symphony/bdk/core/config/model/bdk_rsa_key_config.py
```diff
@@ -0,0 +1,36 @@
+class BdkRsaKeyConfig:
+
+    def __init__(self, path=None, **kwargs):
+        self._path = path
+        self._content = ""
```
symphony-elias, line 5: You could put content as a constructor param with "" as default value.

symphony/bdk/core/config/model/bdk_server_config.py
```diff
@@ -0,0 +1,78 @@
+class BdkServerConfig:
+    _DEFAULT_SCHEME: str = "https"
+    _DEFAULT_HTTPS_PORT: int = 443
+
+    def __init__(self, scheme=None, port=None, host=None, context=None, **kwargs):
+        self._scheme = scheme if scheme is not None else self._DEFAULT_SCHEME
```
symphony-elias: You can put _DEFAULT_SCHEME and _DEFAULT_HTTPS_PORT directly as parameter default values.

symphony/bdk/core/config/model/bdk_ssl_config.py
```diff
@@ -0,0 +1,10 @@
+from symphony.bdk.core.config.model.bdk_certificate_config import BdkCertificateConfig
+
+
+class BdkSslConfig:
+
+    def __init__(self, config):
+        if config is not None:
+            self.trust_store = BdkCertificateConfig(**config.get("trustStore")) if "trustStore" in config else BdkCertificateConfig()
```
symphony-elias: Please put a trustStore param in the constructor, we don't need the full config dict.

symphony/bdk/core/config/bdk_config_loader.py
```diff
@@ -0,0 +1,40 @@
+import json
+from pathlib import Path
+
+from symphony.bdk.core.config.bdk_config_parser import BdkConfigParser
+from symphony.bdk.core.config.exception.bdk_config_exception import BdkConfigException
+from symphony.bdk.core.config.model.bdk_config import BdkConfig
+
+
+class BdkConfigLoader:
```
thibauult, line 9: All classes should be documented

tests/core/config/test_bdk_config_loader.py
```diff
@@ -0,0 +1,83 @@
...
+    assert config.bot.username == "youbot"
+
+
+def test_load_from_file_not_found(wrong_path):
+    fail_error_message = f"Config file has not been found at: {Path(wrong_path).absolute()}"
+    with pytest.raises(BdkConfigException, match=fail_error_message):
+        config = BdkConfigLoader.load_from_file(wrong_path)
+
+
+def test_main_test():
```
symphony-elias: Please remove this test

tests/core/config/test_bdk_config_loader.py
```diff
@@ -0,0 +1,83 @@
...
+    tmp_config_filename = str(uuid.uuid4()) + "-config.yaml"
+    tmp_config_path = Path.home() / ".symphony" / tmp_config_filename
+    tmp_config_path.touch(exist_ok=True)
+    resource_config_content = Path(simple_config_path).read_text()
+    tmp_config_path.write_text(resource_config_content)
+    config = BdkConfigLoader.load_from_symphony_dir(tmp_config_filename)
+    assert config.bot.username == "youbot"
+
+
+def test_load_client_global_config(global_config_path):
```
symphony-elias, line 56: Please test get_base_path returns the correct output when we have a configuration at root level and when we don't have it. Or as suggested, drop the override logic with no scheme, host, port, context at root level.

symphony/bdk/core/config/bdk_config_loader.py
```diff
@@ -0,0 +1,56 @@
...
+    """ Config loader class
+
+    Provide methods to load the a JSON or YAML configuration
```
symphony-youri: "the a"

symphony/bdk/core/config/bdk_config_loader.py
```diff
@@ -0,0 +1,56 @@
...
+    from an absolute path or `USER_HOME_PATH/.symphony``
```
symphony-youri: $HOME is common to point to the home folder (or ~/.symphony)

symphony/bdk/core/config/bdk_config_loader.py
```diff
@@ -0,0 +1,56 @@
...
+    directory or string object to a Python object.
+    """
+
+    @classmethod
+    def load_from_file(cls, config_path: str):
```
symphony-youri: should we use python type hints in our public api (and every where maybe?)

symphony/bdk/core/config/bdk_config_loader.py
```diff
@@ -0,0 +1,56 @@
...
+        """
+        config_path = Path(config_path)
+        if config_path.exists():
+            config_content = config_path.read_text()
+            return cls.load_from_content(config_content)
+        else:
+            raise BdkConfigException(f"Config file has not been found at: {config_path.absolute()}")
+
+    @classmethod
+    def load_from_content(cls, content: str) -> BdkConfig:
```
symphony-youri, line 32: oh there we use it

symphony/bdk/core/config/model/bdk_server_config.py
```diff
@@ -0,0 +1,80 @@
...
+    def get_formated_context(self) -> str:
+        """Formats the context field
+
+        :return: "/" + context if the context is not empty
+        :rtype str
+        """
+        local_context = self.context
+        if local_context == None:
+            return ""
+        if local_context and local_context[0] != "/":
```
symphony-youri, line 30: is it safe to access [0] without checks?

symphony-youness: I don't think it will work if the context is an int ... 
I will add a condition to check if local_context is a string.

symphony/bdk/core/config/model/bdk_config.py
symphony-youness: After thinking about it, (And please tell me if I overthought this one) but what if the dictionary looks like this: (removed c value)
```python
d = {'a': "value a", 'b': "value b", 'd': 'toto'}
```
This will raise the error: `TypeError: _private() missing 1 required positional argument: 'c'`

<br>

I don't think this is a use case we want. This means if we write our code like this:
```python
class BdkConfig(...)
def __init__(self, host, context, port, agent, pod, ...,  **kwargs):
```

and don't provide the `context` at the root of the config file such as the minimalist one:
```yaml
host: acme.symphony.com                                    
bot: 
    username: bot-username                                  
    privateKey:
      path: /path/to/bot/rsa-private-key.pem
```
We will get this error, `TypeError: __init__() missing 1 required positional argument: 'context'` which is not something we want, right ?

docs/configuration.md
```diff
@@ -0,0 +1,49 @@
...
+
+    @staticmethod
+    def run():
+        #
+        config_1 = BdkConfigLoader.load_from_file("/absolute/path/to/config.yaml") #1
+
+        with open("/absolute/path/to/config.yaml") as config_file:                 #2
+            config_2 = BdkConfigLoader.load_from_content(config_file.read())
+
+        config_3 = BdkConfigLoader.load_from_symphony_dir("/absolute/path/to/config.yaml") #3
```
thibauult: Param should be a relative location from the `~/.symphony` directory

examples/core/config/config_main.py
```diff
@@ -0,0 +1,20 @@
...
+class ConfigMain:
+
+    @staticmethod
+    def run():
+        config_1 = BdkConfigLoader.load_from_file("/absolute/path/to/config.yaml")
+
+        with open("/absolute/path/to/config.yaml") as config_file:
+            config_2 = BdkConfigLoader.load_from_content(config_file.read())
+
+        config_3 = BdkConfigLoader.load_from_symphony_dir("/absolute/path/to/config.yaml")
```
thibauult: Same comment, path is wrong here

tests/resources/config/config.json
```diff
@@ -0,0 +1,17 @@
...
+  },
+  "agent": {
+    "host": "devx1.symphony.com"
+  },
+  "keyManager": {
+    "host": "devx1.symphony.com"
+  },
+  "bot": {
+    "username": "youbot",
+    "privateKeyPath": "/Users/local/conf/agent/privatekey.pem"
```
thibauult: `bot` > `privateKey` > `path`

tests/resources/config/config_global.json
```diff
@@ -0,0 +1,28 @@
...
+  "keyManager": {
+    "host": "devx1.symphony.com",
+    "context": "diff-context"
+  },
+  "sessionAuth": {
+    "scheme": "http"
+  },
+  "bot": {
+    "username": "tibot",
+    "privateKeyPath": "/Users/local/conf/agent/privatekey.pem"
```
thibauult: Python BDK should not support `privateKeyPath` (it's something already deprecated in Java), but `privateKey` > `path` instead.

symphony/bdk/core/config/model/bdk_server_config.py
symphony-youness: I can't do that because we are passing arguments using `dict.get()` that is passing `None` as an argument which overrides the default parameters.
```python
>>> BdkConfig
 config = {}
super().__init__(scheme=config.get("scheme"))

#Which overrides the default value in 
>>> BdkServerConfig 
def __init__(self, scheme=None)
```

symphony/bdk/core/config/model/bdk_config.py
symphony-elias: What about `context = None` in the parameters ?

tests/resources/config/config_global.json
symphony-youness: this was a mistake. it should not be used.

pyproject.toml
```diff
@@ -29,7 +30,6 @@ junit_logging = "all"
 [tool.coverage.run]
 branch = true
 source = ["symphony"]
-omit = ["symphony/bdk/gen/*"]
```
symphony-elias: Please keep that as we don't want to run coverage checks on the generated code.

docs/configuration.md
```diff
@@ -0,0 +1,49 @@
...
+        with open("/absolute/path/to/config.yaml") as config_file:                 #2
+            config_2 = BdkConfigLoader.load_from_content(config_file.read())
+
+        config_3 = BdkConfigLoader.load_from_symphony_dir("config.yaml") #3
+
+        config_4 = BdkConfig() #4
+
+if __name__ == "__main__":
+    ConfigMain.run()
+```
```
thibauult: Could we make doc examples a bit simpler: 
```python
from symphony.bdk.core.config.bdk_config_loader import BdkConfigLoader
from symphony.bdk.core.config.model.bdk_config import BdkConfig

if __name__ == "__main__":

  config_1 = BdkConfigLoader.load_from_file("/absolute/path/to/config.yaml") #1

  with open("/absolute/path/to/config.yaml") as config_file:                 #2
    config_2 = BdkConfigLoader.load_from_content(config_file.read())

  config_3 = BdkConfigLoader.load_from_symphony_dir("config.yaml") #3

  config_4 = BdkConfig() #4
```

Let's also try to improve imports. For instance, could we only have: 
```python
from symphony.bdk.core.config import BdkConfigLoader, BdkConfig
```
?

symphony-youness: For the imports. directly like this, no we can't. I will check if we can do smth to fix that.<br>
For the doc we can definitly do this as we don't need a main in the example and it works
```python
config_1 = BdkConfigLoader.load_from_file("/absolute/path/to/config.yaml") #1
config_2 = BdkConfigLoader.load_from_symphony_dir("relative/path/to/config.yaml") #2
config_3 = BdkConfig() #3
```

docs/configuration.md
```diff
@@ -0,0 +1,49 @@
...
+```python
+from symphony.bdk.core.config.bdk_config_loader import BdkConfigLoader
+from symphony.bdk.core.config.model.bdk_config import BdkConfig
+
+
+class ConfigMain:
+
+    @staticmethod
+    def run():
+        #
```
symphony-elias: Remove the useless hash

symphony/bdk/core/config/model/bdk_certificate_config.py
```diff
@@ -0,0 +1,22 @@
...
+        """
+        return (self.path is not None or self.content != "") and self.password is not None
+
+    def is_valid(self) -> bool:
+        """Check if the certificate configuration is valid.
+        If both certificate path and content, the configuration is invalid.
+
+        :return: true if the RSA configuration is invalid.
+        """
+        return (self.path is not None and self.content != "") and self.password is not None
```
symphony-elias: I guess it is: `(self.path is not None *or* self.content != "") and self.password is not None`, i.e. you have either a path to a cert file or the cert content.

symphony/bdk/core/config/model/bdk_config.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class BdkConfig(BdkServerConfig):
+    """Class containing the Bdk configuration
+
+    :param **config: dict Configuration parameters
+    :return self: BdkConfig instance
+    """
+    def __init__(self, **config):
+        super().__init__(scheme=config.get("scheme"), host=config.get("host"), port=config.get("port"),
```
symphony-elias, line 14: Please do the same as BdkCertificateConfig constructor.

symphony/bdk/core/config/model/bdk_certificate_config.py
symphony-youness: ah I think I misunderstood that one then.
it's probably 
```python
not (self.path is not None and self.content != "") and self.password is not None
```
to enforce only one being used. <br>
In that aspect I suppose, I should do the same for the rsa private key.

symphony/bdk/core/config/model/bdk_config.py
symphony-youness: ok

symphony/bdk/core/config/model/bdk_certificate_config.py
```diff
@@ -0,0 +1,22 @@
...
+
+        :return: true if the certificate authentication is configured
+        """
+        return (self.path is not None or self.content != "") and self.password is not None
+
+    def is_valid(self) -> bool:
+        """Check if the certificate configuration is valid.
+        If both certificate path and content, the configuration is invalid.
+
+        :return: true if the RSA configuration is invalid.
```
symphony-elias: `if the RSA configuration is invalid` -> should be `valid` instead.

symphony/bdk/core/config/model/bdk_rsa_key_config.py
```diff
@@ -0,0 +1,21 @@
...
+
+        :return: true if the RSA authentication is configured
+        """
+        return self.path is not None or self.content != ""
+
+    def is_valid(self) -> bool:
+        """Check if the RSA configuration is valid.
+        If both of private key path and content, the configuration is invalid.
+
+        :return: true if the RSA configuration is invalid.
```
symphony-elias: Should be `true if the RSA configuration is valid`

symphony-elias (approved):

PR #104, opened 2021-01-20
PLAT-10275: RSA authentication implementation

symphony-hong:
**Draft PR without unittest and documentation yet**

### Ticket
[PLAT-10275](https://perzoinc.atlassian.net/browse/PLAT-10275)

### Description
RSA Authentication implementation

### Dependencies

### Checklist


symphony/bdk/core/auth/auth_session.py
```diff
@@ -0,0 +1,34 @@
+class AuthSessionRSA:
+    """
+    RSA Authentication session handle.
+    """
+
+    def __init__(self, authenticator):
+        self.__session_token = None
```
symphony-elias: Please use a single underscore. For the meaning of single and double underscores in python: https://stackoverflow.com/questions/1301346/what-is-the-meaning-of-single-and-double-underscore-before-an-object-name

symphony/bdk/core/auth/auth_session.py
```diff
@@ -0,0 +1,34 @@
...
+        self.__key_manager_token = None
+        self.__authenticator = authenticator
+
+    def get_session_token(self) -> str:
```
symphony-elias: The following would be more pythonic:
```python
@property
def session_token(self)
```

symphony/bdk/core/auth/auth_session.py
```diff
@@ -0,0 +1,34 @@
...
+        """
+        Get the pod's authentication token.
+
+        Returns: the Pod session token.
```
symphony-elias: Please use [reStructuredTex](https://docutils.sourceforge.io/rst.html) as we will use Sphynx to generate documentation. It is also the standard advocated by [PEP-287](https://www.python.org/dev/peps/pep-0287/)

symphony/bdk/core/auth/authenticator_factory.py
```diff
@@ -0,0 +1,34 @@
...
+
+    def get_bot_authenticator(self) -> BotAuthenticator:
+        """
+        Creates a new instance of a Bot Authenticator service.
+
+        Returns: a new BotAuthenticator instance.
+
+        """
+        if self.__config.bot.is_rsa_authentication_configured:
+            if not self.__config.bot.is_rsa_configuration_valid:
```
symphony-elias: I believe it is with the parentheses : `self.__config.bot.is_rsa_configuration_valid()`.

symphony/bdk/core/auth/bot_authenticator.py
```diff
@@ -0,0 +1,78 @@
...
+from symphony.bdk.core.auth.exception.bdk_authentication_exception import AuthUnauthorizedException
+from symphony.bdk.core.config.model.bdk_bot_config import BdkBotConfig
+
+
+class BotAuthenticator(ABC):
+    """
+    Bot authentication service.
+    """
+
+    async def _retrieve_token(self, api_client: ApiClient):
```
symphony-elias: Please provide a type annotation about the returned type

symphony/bdk/core/auth/bot_authenticator.py
```diff
@@ -0,0 +1,78 @@
...
+        unauthorized_message = "Service account is not authorized to authenticate. Check if credentials are valid."
+        jwt = create_signed_jwt(self.bot_config.private_key, self.bot_config.username)
+        req = AuthenticateRequest(token=jwt)
+        try:
+            token = await AuthenticationApi(api_client).pubkey_authenticate_post(req)
+            return token.token
+        except ApiException:
+            raise AuthUnauthorizedException(unauthorized_message)
+        finally:
+            await api_client.close()
```
symphony-elias: Do we assume we pass a new `api_client` for each call ? Or do we want to reuse `api_client` across several calls (if possible) ?

symphony/bdk/core/auth/auth_session.py
```diff
@@ -0,0 +1,34 @@
+class AuthSessionRSA:
```
symphony-elias: Not sure this is specific to RSA authentication. `AuthSession` should be fine.

symphony/bdk/core/auth/bot_authenticator.py
symphony-hong: I think the `api_client` don't need to be closed manually at all as it suggested in the legacy sdk: https://github.com/SymphonyPlatformSolutions/symphony-api-client-python/blob/2d4e3c5f24c6cc464254b4210fe516323fffd6e4/sym_api_client_python/clients/sym_bot_client.py#L349-L368
WDYT?

symphony-elias: We can afford not to close api_client objects if we keep during the whole execution of the bot. However, I got this kind of error in my PoC when not closing the clients:
```
Unclosed client session
client_session: <aiohttp.client.ClientSession object at 0x10597f100>
Unclosed connector
connections: ['[(<aiohttp.client_proto.ResponseHandler object at 0x105979400>, 2.0372855)]']
connector: <aiohttp.connector.TCPConnector object at 0x105377c40>
```
Please check if there is a clean way to do so when the asyncio loop is terminating

symphony-hong: Once we close an api client, we have to pass a new instance to make another call and aiohttp doc does not recommend doing this way.
But I'm pretty stuck on how we can close the api_client in a clean way. Maybe we can define a method at the bdk level to close all the existing api clients and let user close them manually?
And actually that's not the error from aiohttp but only a warning I think: https://github.com/aio-libs/aiohttp/blob/3be8a68d750fd9902106e747d11a0622b8650f2c/aiohttp/client.py#L286

symphony-hong: @symphony-youness do you have any idea for this issue?

symphony-youness: As @symphony-hong said, they were warnings. Because the session was not closed at the end. So we won't have that warning until we terminate the program.

However, we might then want to use `.close()` in a method to gracefully shutdown the bot

symphony-youness: I think that's how it was used in the legacy one.
When we have an exception that we didn't handle during a `read_datafeed`, the `.close` function is called in `DatafeedEventService.deactivate_datafeed()` to gracefully stop the bot.

symphony-elias: I didn't notice it was only a warning. So let's keep it that way and reuse the same api_client without closing

symphony/bdk/core/auth/jwt_helper.py
```diff
@@ -0,0 +1,29 @@
...
+    expiration = expiration if expiration is not None else int(
+        datetime.datetime.now(datetime.timezone.utc).timestamp() + (5 * 58))
+    payload = {
+        "sub": username,
+        "exp": expiration
+    }
+    return jwt.encode(payload, private_key, algorithm="RS512")
+
+
+def __get_key(private_key_path):
```
symphony-elias: Please replace double underscore with single underscore. Single underscore is the python standard for private fields and methods.

symphony/bdk/core/auth/auth_session.py
```diff
@@ -0,0 +1,30 @@
+class AuthSession:
+    """RSA Authentication session handle.
+    """
+
+    def __init__(self, authenticator):
+        self._session_token = None
+        self._key_manager_token = None
+        self._authenticator = authenticator
```
symphony-youness, line 8: @symphony-elias Please correct me if I am wrong.

We decided to not use the underscore for 'private' attributes. 

For a **single underscore** is merely an agreed-upon convention and does not impose any restrictions on accessing the value of an attribute.

For a **double underscore**, Python mangles these names and it is used to avoid name clashes with names defined by subclasses.

I'll make a page to define these conventions, we only made these choices during the config PR so no worries ^^

symphony/bdk/core/auth/auth_session.py
```diff
@@ -0,0 +1,30 @@
...
+
+    @property
+    def session_token(self) -> str:
```
symphony-youness: Maybe use properties only if we need to change the behavior of accessing a certain function.
So usually we won't use any properties for direct access to an attribute

symphony/bdk/core/auth/auth_session.py
symphony-elias: It depends, we will use single underscore for private attributes if there is a specific logic for getters or setters (in which case, we will use single underscore for the attribute + properties). Or if the attribute is not expected to be read form the outside.

symphony-elias (approved):
LGTM !

pyproject.toml
```diff
@@ -11,11 +11,14 @@ python-dateutil = "^2.8.1"
 urllib3 = "^1.26.2"
 aiohttp = "^3.7.3"
 pyyaml = "^5.3.1"
```
symphony-youri, line 13: has pyjwt been considered? (it seems more popular on GH)

symphony-youri (commented):
Is pylint configured to fail the build?  I see warnings when I run it
and it is configured with fail-under=0.00

symphony/bdk/core/auth/jwt_helper.py
```diff
@@ -0,0 +1,29 @@
...
+    :param private_key_config: The private key configuration for a service account or an extension app.
+    :param username: the username of the user to authenticate
+    :param expiration: expiration of the authentication request in seconds.
+
+    :return: a signed JWT for a specific user or an extension app.
+    """
+    private_key = _get_key(private_key_config.path) \
+        if private_key_config.path is not None else private_key_config.content
+    expiration = expiration if expiration is not None else int(
+        datetime.datetime.now(datetime.timezone.utc).timestamp() + (5 * 58))
```
symphony-youri, line 17: 5 * 58 ? why?

symphony-hong: I took this value from the python legacy sdk: https://github.com/SymphonyPlatformSolutions/symphony-api-client-python/blob/master/sym_api_client_python/auth/rsa_auth.py#L76-L77

symphony-youri (commented):
Will we implement tests mocking a real/mocked server (i.e using JSON responses) or just pure unit tests with object mocks?

symphony-hong:
> Is pylint configured to fail the build? I see warnings when I run it

it seems the pylint won't fail the build.

> Will we implement tests mocking a real/mocked server (i.e using JSON responses) or just pure unit tests with object mocks?

I think we should consider to use a mocked server but I haven't tried to build a mocked server yet

pyproject.toml
symphony-hong: jose has been used in the legacy and the Elias's POC, so I chose it subjectively. I will have a look at pyjwt and see if we should use it instead.

PR #107, opened 2021-01-28
PLAT-10367: Message Service implementation.

symphony-hong:
### Ticket
[PLAT-10367](https://perzoinc.atlassian.net/browse/PLAT-10367)

### Description
Message Service implementation

### Dependencies

### Checklist


symphony/bdk/core/symphony_bdk.py
```diff
@@ -26,6 +37,20 @@ async def bot_session(self) -> AuthSession:
...
+        if self._message_service is None:
+            self._message_service = MessageService(
+                MultiAttachmentsMessagesApi(self._agent_client),
+                MessageApi(self._pod_client),
+                MessageSuppressionApi(self._pod_client),
+                StreamsApi(self._pod_client),
+                PodApi(self._pod_client),
+                AttachmentsApi(self._agent_client),
+                DefaultApi(self._pod_client),
+                await self.bot_session()
```
symphony-youness: Regarding the lazy init, I discussed it quickly with @symphony-elias and this works fine for now !!

symphony/bdk/core/service/message_service.py
```diff
@@ -0,0 +1,366 @@
...
+                    'multipart/form-data'
+                ]
+            },
+            api_client=api_client,
+            callable=__v4_message_create_post
+        )
+
+
+class MessageService:
+    def __init__(self, messages_api: MultiAttachmentsMessagesApi,
```
symphony-youness: Just to check. I am having some confusion. 
What problem does the `MultiAttachmentsMessagesApi` solve that the generated `MessagesApi` has? I can't seem to find it on the java BDK.
Is it related to some new specs?

symphony-hong: The generated `MessagesApi` doesn't allow to send a message with multiple attachments which is allowed by the api provided by the agent. That's why I create the `MultiAttachmentsMessagesApi` which accepts the attachment field as a list of file.

symphony/bdk/core/service/message_service.py
```diff
@@ -0,0 +1,366 @@
...
+
+    async def get_message(self, message_id: str) -> V4Message:
+        params = {
+            'id': message_id,
+            'session_token': self._auth_session.session_token,
+            'key_manager_token': self._auth_session.key_manager_token
+        }
+        return await self._messages_api.v1_message_id_get(**params)
+
+    async def list_attachments(self, stream_id: str, **kwargs) -> [StreamAttachmentItem]:
```
symphony-youri: is the kwargs temporary? if we use it don't we loose completion/documentation?

symphony/bdk/core/service/message_service.py
```diff
@@ -0,0 +1,366 @@
...
+from symphony.bdk.gen.pod_api.pod_api import PodApi
+from symphony.bdk.gen.pod_api.streams_api import StreamsApi
+from symphony.bdk.gen.pod_model.message_metadata_response import MessageMetadataResponse
+from symphony.bdk.gen.pod_model.message_receipt_detail_response import MessageReceiptDetailResponse
+from symphony.bdk.gen.pod_model.message_status import MessageStatus
+from symphony.bdk.gen.pod_model.message_suppression_response import MessageSuppressionResponse
+from symphony.bdk.gen.pod_model.stream_attachment_item import StreamAttachmentItem
+
+
+class MultiAttachmentsMessagesApi(MessagesApi):
```
symphony-youri: is it ok to put both classes in the same file? to keep it simple would not it be simpler to have a separate file?

symphony/bdk/core/service/message_service.py
```diff
@@ -0,0 +1,366 @@
...
+        self.v4_multi_attachment_message_blast_post = Endpoint(
+            settings={
+                'response_type': (V4MessageBlastResponse,),
+                'auth': [],
+                'endpoint_path': '/v4/message/blast',
+                'operation_id': 'v4_message_blast_post',
+                'http_method': 'POST',
+                'servers': None,
+            },
+            params_map={
```
symphony-youri: looks like from here it is similar to above, maybe we can factorize it?

symphony/bdk/core/service/message_service.py
symphony-hong: I don't really understand your point that the kwargs temporary. IMHO, the **kwargs contains all the optional fields that we gonna pass to the api and apparently we have to document all the fields that we can pass to the api through kwargs

symphony/bdk/core/service/message_service.py
symphony-hong: The file containing the MessagesApi is generated from the specs and we shouldn't modify it, that's why I keep this class here.

symphony-youri: and we cannot move it to its own file?

symphony/bdk/core/service/message_service.py
symphony-youri: and what about using default arguments? (named ones like here https://www.geeksforgeeks.org/default-arguments-in-python/)
could that help to be more explicit?

symphony/bdk/core/service/message_service.py
symphony-hong: actually we have a slight difference between `sid` and `sids`

symphony/bdk/core/service/message/message_service.py
```diff
@@ -0,0 +1,217 @@
...
+        params = {
+            'sid': stream_id,
+            'session_token': self._auth_session.session_token,
+            'key_manager_token': self._auth_session.key_manager_token,
+            'message': message,
+            'data': data,
+            'version': version
+        }
+        if attachment is not None:
+            params['attachment'] = attachment
```
symphony-elias: I'd prefer to have everything related to `attachment` in a single place like: remove line 70 and have:
```
if attachment is not None:
    params['attachment'] = attachment if type(attachment) is list else [attachment]
```
Same for `preview`.

symphony/bdk/core/service/message/multi_attachments_messages_api.py
```diff
@@ -0,0 +1,214 @@
...
+)
+
+
+class MultiAttachmentsMessagesApi(MessagesApi):
+
+    def __init__(self, api_client=None):
+        super().__init__(api_client)
+
+        def __v4_message_create_post(self, **kwargs):
+            kwargs['async_req'] = kwargs.get(
```
symphony-elias, line 19: Could you explain why this is needed? I was expecting the default values to be what you put here.

symphony/bdk/core/service/message/multi_attachments_messages_api.py
```diff
@@ -0,0 +1,214 @@
...
+            kwargs['_request_timeout'] = kwargs.get(
+                '_request_timeout', None
+            )
+            kwargs['_check_input_type'] = kwargs.get(
+                '_check_input_type', True
+            )
+            kwargs['_check_return_type'] = kwargs.get(
+                '_check_return_type', True
+            )
+            kwargs['_host_index'] = kwargs.get('_host_index')
```
symphony-elias, line 37: Please mind that it only puts a None value for key `_host_index` if it key was not existing before. Is this needed?

symphony/bdk/core/service/message/multi_attachments_messages_api.py
symphony-hong: Actually I took inspiration from what was generated in the `MessagesApi` class, I refactored a bit but I decided to keep the fields that I'm not sure to be safe: https://github.com/SymphonyPlatformSolutions/symphony-api-client-python/blob/2.0/symphony/bdk/gen/agent_api/messages_api.py#L859-L877

symphony/bdk/core/service/message/multi_attachments_messages_api.py
symphony-hong: same answer as the comment above

symphony/bdk/core/service/message/multi_attachments_messages_api.py
symphony-elias: 👌 for me

symphony/bdk/core/service/message/multi_attachments_messages_api.py
symphony-elias: 👌

symphony-elias (approved):
Looks good !

symphony-elias:
Please add some documentation in the docs/ folder, at least explain how to use the message service.

symphony/bdk/core/symphony_bdk.py
```diff
@@ -26,6 +37,26 @@ async def bot_session(self) -> AuthSession:
             self._bot_session = await self._authenticator_factory.get_bot_authenticator().authenticate_bot()
         return self._bot_session
 
+    async def messages(self):
```
symphony-youri: is it ok to make it async?

For instance should we perform any remote call when creating the message service?
Or should we do the first calls only when actually using the service for the first time?

symphony-youri: Should we provide the returned type here?

symphony-youri: (with the -> in the signature)

docs/message_service.md
```diff
@@ -0,0 +1,47 @@
...
+            with open("/path/to/attachment1", "rb") as file1, \
+                    open("/path/to/attachment2", "rb") as file2:
+                await message_service.send_message(
+                    stream_id_1,
+                    "<messageML>Hello, World!</messageML>",
+                    attachment=[file1, file2]
+                )
+
+        finally:
+            await bdk.close_clients()
```
symphony-youri: should we use a with: block (or a sort of automatic close)? (to avoid forgetting about closing the clients)

symphony/bdk/core/symphony_bdk.py
symphony-hong: the problem here is we need a bot session to initialise the message service instance. So I guess, we should pass the `authenticator` to the message service constructor then inside the message service, each time we make a api call we gonna check the bot session there?

symphony/bdk/core/service/message/message_service.py
```diff
@@ -0,0 +1,340 @@
...
+        """
+        params = {
+            'sid': stream_id,
+            'since': since,
+            'session_token': self._auth_session.session_token,
+            'key_manager_token': self._auth_session.key_manager_token,
+            'skip': skip,
+            'limit': limit
+        }
+        message_list = await self._messages_api.v4_stream_sid_message_get(**params)
```
symphony-elias, line 73: What about `self._messages_api.v4_stream_sid_message_get(sid=stream_id, since=since, session_token=self._auth_session.session_token, ...)`
IMHO having a param dict which is then unpacked makes things more complicated.

symphony-hong: I would leave it as a param dict because for instance, `messages_api.v4_stream_sid_message_get` is defined as an instance of `Endpoint` callable class accepting the **kwargs parameters [Endpoint](https://github.com/SymphonyPlatformSolutions/symphony-api-client-python/blob/2.0/symphony/bdk/gen/api_client.py#L751-L762)
And IMHO, it doesn't impact much to the complexity of the code

symphony/bdk/core/symphony_bdk.py
symphony-hong: I think it's okay to leave this method async. In fact, it's only really async at the first time we authenticate the bot to get the `bot_session`.
We can also init the bot session as `AuthSession(bot_authenticator)`, but then inside the message, at the beginning of each service method, we have to check if the token is None or not, then authenticate the bot, and TBH I don't really like this solution.
WDYT?

symphony-youri: What if the check is part of the getter in AuthSession?
One problem I see is that in case of concurrent access we might authenticate twice (if we don't lock) but with async it is not going to happen anyway

symphony/bdk/core/service/message/message_service.py
symphony-elias: I'd like to have someone else's opinion on that one, like @symphony-youri

symphony-youri: no strong opinion on that because it is internal and not exposed to users (and I have tried to quickly search for guidelines on how to be idiomatic on this and did not find anything to help me decide)

We have other places where add parameters conditionally (for attachments) so maybe we still need a map?
I would perhaps have a small method to reuse to avoid repeating the session_token part everywhere?

symphony/bdk/core/auth/auth_session.py
```diff
@@ -3,12 +3,32 @@ class AuthSession:
...
     async def refresh(self):
         """Trigger re-authentication to refresh the tokens.
         """
-        self.session_token = await self._authenticator.retrieve_session_token()
-        self.key_manager_token = await self._authenticator.retrieve_key_manager_token()
+        self._session_token = await self._authenticator.retrieve_session_token()
+        self._key_manager_token = await self._authenticator.retrieve_key_manager_token()
+
+    @property
+    async def session_token(self):
```
symphony-youri, line 17: and now when I see as async I'm wondering if it is the right way to go (I'm lacking experience on async in Python here :)) 
let's go this way for now and we can always reevaluate that later

symphony-youri (approved):

PR #108, opened 2021-02-01
PLAT-10342 Datafeed service

symphony-youness:
### Ticket
[PLAT-10342](https://perzoinc.atlassian.net/browse/PLAT-10342)

### Description
Datafeed service implementation

Another ticket will be implementing 

### Checklist


symphony/bdk/core/datafeed/abstract_datafeed_loop.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def subscribe(self, listener):
+        self.listeners.append(listener)
+
+    def unsubscribe(self, listener):
+        self.listeners.remove(listener)
+
+    def handle_v4_event_list(self, events: List[V4Event]):
+        for event in events:
+            if event.type is None:  # should we ever check this ? Isn't the generated code supposed to check for nullity or not the right type and raise an exception
```
symphony-elias: Indeed, I don't know if we can get an empty event type from the read DF endpoint.

symphony/bdk/core/datafeed/abstract_datafeed_loop.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def dispatch_on_event_type(self, listener: RealTimeEventListener, event: V4Event):
+        try:
+            listener_method_name, payload_field_name = self.dispatch_dict[event.type]
+            listener_method = getattr(listener, listener_method_name)
+            listener_method(event.initiator, getattr(event.payload, payload_field_name))
+        except KeyError:
+            print(f"Received event with an unknown type: {event.type}")
+
+    def handle_event(self, event: V4Event):
```
symphony-elias: I guess you can remove that

symphony/bdk/core/datafeed/datafeed_loop_v1.py
```diff
@@ -0,0 +1,48 @@
...
+        self.started = False
+
+    async def start(self):
+        try:
+            if self.datafeed_id == "":
+                self.datafeed_id = await self.create_datafeed_and_persist()
+            self.started = True
+            while self.started:  # I would like to know why do we use a do while in the java bdk for the read datafeed ?
+                await self.read_datafeed()
+        except ApiException:
```
symphony-elias: You should filter on HTTP 400 and this should be done inside the while loop.

symphony/bdk/core/datafeed/datafeed_loop_v1.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def retrieve_datafeed(self):
+        return self.datafeed_repository.read()
+
+    async def create_datafeed_and_persist(self):
+        params = {
+            'session_token': self.auth_session.session_token,
+            'key_manager_token': self.auth_session.key_manager_token,
+        }
+        response = await self.datafeed_api.v4_datafeed_create_post(**params)
```
symphony-elias: Please remove the `params` dict and put directly the params like: `self.datafeed_api.v4_datafeed_create_post(session_token=self.auth_session.session_token, key_manager_token=self.auth_session.key_manager_token)`

symphony/bdk/core/datafeed/on_disk_datafeed_id_repository.py
```diff
@@ -0,0 +1,80 @@
...
+    DATAFEED_ID_FILE = "datafeed.id"
+
+    def __init__(self, config: BdkConfig):
+        self.config = config
+
+    def write(self, datafeed_id, agent_base_path=""):
+        content = datafeed_id + "@" + agent_base_path
+        try:
+            self._get_datafeed_id_file().write_text(content, encoding='utf-8')
+        except ValueError:
```
symphony-elias: Why `ValueError`? What about I/O error (like file system permission issues)?

symphony/bdk/core/datafeed/real_time_event_listener.py
```diff
@@ -0,0 +1,83 @@
...
+from symphony.bdk.gen.agent_model.v4_user_left_room import V4UserLeftRoom
+from symphony.bdk.gen.agent_model.v4_room_member_promoted_to_owner import V4RoomMemberPromotedToOwner
+from symphony.bdk.gen.agent_model.v4_room_member_demoted_from_owner import V4RoomMemberDemotedFromOwner
+from symphony.bdk.gen.agent_model.v4_connection_requested import V4ConnectionRequested
+from symphony.bdk.gen.agent_model.v4_connection_accepted import V4ConnectionAccepted
+from symphony.bdk.gen.agent_model.v4_message_suppressed import V4MessageSuppressed
+from symphony.bdk.gen.agent_model.v4_symphony_elements_action import V4SymphonyElementsAction
+
+
+class RealTimeEventListener:
```
symphony-elias: I believe this should be an ABC ?

symphony/bdk/core/datafeed/on_disk_datafeed_id_repository.py
symphony-youness: This was for handling encoding errors but the content shouldn't be other than `utf-8`.
**I will remove it**

symphony/bdk/core/datafeed/on_disk_datafeed_id_repository.py
```diff
@@ -0,0 +1,80 @@
...
+    @abstractmethod
+    def read(self):
+        """Reads the persisted datafeed from storage.
+
+        :return the persisted datafeed id if present in storage, an empty string otherwise
+        """
+        pass
+
+    @abstractmethod
+    def read_agent_base_path(self):
```
symphony-youri: should we add the returned types?

symphony/bdk/core/datafeed/real_time_event_listener.py
symphony-youness: The user might want to implement some listeners and not all of them right? 
This is why I didn't use the @abstractmethod decorator to force the implementation of the methods.
Why make it an ABC then ?

symphony/bdk/core/datafeed/datafeed_loop_v1.py
```diff
@@ -0,0 +1,48 @@
...
+        self.datafeed_repository = OnDiskDatafeedIdRepository(config) if repository is None else repository
+        self.datafeed_id = self.retrieve_datafeed()
+        self.started = False
+
+    async def start(self):
+        try:
+            if self.datafeed_id == "":
+                self.datafeed_id = await self.create_datafeed_and_persist()
+            self.started = True
+            while self.started:  # I would like to know why do we use a do while in the java bdk for the read datafeed ?
```
symphony-youri: to set the started status the first time?

symphony/bdk/core/datafeed/datafeed_loop_v1.py
```diff
@@ -0,0 +1,48 @@
...
+        events = await self.datafeed_api.v4_datafeed_id_read_get(self.datafeed_id,
+                                                                 self.auth_session.session_token,
+                                                                 self.auth_session.key_manager_token)
+        if events is not None and not (events.value == []):  # need to check if the call ever return None
+            self.handle_v4_event_list(events.value)
+
+    def retrieve_datafeed(self):
+        return self.datafeed_repository.read()
+
+    async def create_datafeed_and_persist(self):
```
symphony-youri: should we mark these methods as private?

symphony/bdk/core/datafeed/abstract_datafeed_loop.py
```diff
@@ -0,0 +1,93 @@
...
+
+class AbstractDatafeedLoop(DatafeedLoop, ABC):
+    def __init__(self, datafeed_api: DatafeedApi, auth_session: AuthSession, config: BdkConfig):
+        self.datafeed_api = datafeed_api
+        self.listeners = []
+        self.auth_session = auth_session
+        self.bdk_config = config
+        self.api_client = datafeed_api.api_client
+        self.dispatch_dict = {
+            'MESSAGESENT': ('on_message_sent', 'message_sent'),
```
symphony-youri: is there a way in Python we could use method reference directly instead of strings?

symphony-youri: Something like https://stackoverflow.com/questions/1351669/referencing-class-methods-in-class-lists-in-python maybe?

symphony/bdk/core/datafeed/abstract_datafeed_loop.py
symphony-youness: My understanding from the generated code in the `V4Event` class is that:
- whatever value the type is, if present, it's going to be returned.
- if no type is present in the response, then there no attribute `event.type` will be created. Which will raise an exception.
-> I will handle it using a try except block in the `dispatch_on_event_type` function

symphony/bdk/core/datafeed/datafeed_loop_v1.py
symphony-youness: Please tell me if I am wrong but we don't really get a response if the status code is 400, the generated code always throws an exception. **I will look more into this and how to get the status code.**
<br>
I tried to handle it like this, but if there is a problem with the read_datafeed endpoint and not the create datafeed endpoint, we end up in an endless loop.
```python
    async def start(self):
        if self.datafeed_id == "":
            self.datafeed_id = await self.create_datafeed_and_persist()
        self.started = True
        while self.started:
            try:
                await self.read_datafeed()
            except ApiException:
                try:
                    self.datafeed_id = await self.create_datafeed_and_persist()
                except ApiException:
                    raise
```

symphony/bdk/core/config/model/bdk_datafeed_config.py
```diff
@@ -0,0 +1,17 @@
+from pathlib import Path
+
+
+class BdkDatafeedConfig:
+
+    def __init__(self, config):
+        self.id_file_path = ""
```
symphony-youri, line 7: should we read it from the config?

symphony/bdk/core/datafeed/abstract_datafeed_loop.py
symphony-youness: I spent some time on that one, I agree with you it could have been better.

The problem with this is that we want to access the user implementation of the `on_message_sent`function. 
We can do it using `listener.on_message_sent`. But we can't really pass the listener to a dict or an enum even.

A better solution was using decorators (@singledispatch) but I found it a bit complex to implement. So I'll implement it as an improvement, once I figure out how it works.

pyproject.toml
```diff
@@ -33,7 +33,7 @@ junit_logging = "all"
 [tool.coverage.run]
 branch = true
 source = ["symphony"]
-omit = ["symphony/bdk/gen/*"]
+omit = ["symphony/bdk/gen/*", "symphony/bdk/core/datafeed/*"]
```
thibauult: We won't test this feature?

symphony-youness: I will create a ticket with no points to add tests for next sprint. 

This is purely for the build to pass.

symphony/bdk/core/datafeed/datafeed_loop_v1.py
symphony-elias: You can check this way: 
```
except ApiException as e:
  if e.status == 400:
    ...
```

symphony/bdk/core/datafeed/real_time_event_listener.py
symphony-elias: 👌 I get it

symphony/bdk/core/datafeed/datafeed_loop_v1.py
```diff
@@ -12,34 +31,34 @@ def __init__(self, datafeed_api, auth_session, config, repository=None):
...
+                try:
+                    self.datafeed_id = await self.__create_datafeed_and_persist()
+                except ApiException:
+                    raise
 
     def stop(self):
         self.started = False
 
-    async def read_datafeed(self):
+    async def __read_datafeed(self):
```
symphony-elias: Single underscore would be enough.

symphony/bdk/core/datafeed/datafeed_loop_v1.py
```diff
@@ -0,0 +1,62 @@
...
+    async def start(self):
+        if self.datafeed_id == "":
+            self.datafeed_id = await self.__create_datafeed_and_persist()
+        self.started = True
+        while self.started:
+            try:
+                await self.__read_datafeed()
+            except ApiException as e:
+                if e.status == 400:
+                    self.datafeed_id = await self.__create_datafeed_and_persist()
```
symphony-elias: else we raise the `ApiException` because we should only filter HTTP 400 errors.

tests/resources/datafeed/datafeedId
```diff
@@ -0,0 +1 @@
+8e7c8672-220@localhost:7443
```
thibauult, line 1: To be ignored from `.gitignore`

symphony/bdk/core/service/datafeed/on_disk_datafeed_id_repository.py
```diff
@@ -0,0 +1,78 @@
+from abc import ABC, abstractmethod
+from pathlib import Path
+from symphony.bdk.core.config.model.bdk_config import BdkConfig
+
+
+class DatafeedIdRepository(ABC):
```
thibauult, line 6: This file is called `on_disk_datafeed_id_repository.py`, however you define the `DatafeedIdRepository` interface + the `OnDiskDatafeedIdRepository` implem. 

I'd move those classes to `symphony/bdk/core/service/datafeed/__init__.py` to make imports simpler

docs/datafeed.md
```diff
@@ -0,0 +1,21 @@
...
+The datafeed service is a service used for handling the [_Real Time Events_](https://developers.symphony.com/restapi/docs/real-time-events). 
+
+When a user makes an interaction within the IM, MIM or Room chat like sending a message, joining or leaving a room chat..., 
+when a connection request is sent, when a wall post is published or when a user replies an Symphony element, an event will be sent to the datafeed.
+
+The datafeed service is a core service built on top of the Datafeed API and provides a dedicated contract to bot developers to work with datafeed. 
+
+
+## How to use
+The central component for the contract between bot developers and  the Datafeed API is the `DatafeedService`.
```
symphony-youri, line 12: extra space between and..the (yes I'm a maniac)

symphony-youri: DatafeedService -> Datafeed Loop?

symphony/bdk/core/service/datafeed/abstract_datafeed_loop.py
```diff
@@ -84,10 +88,8 @@ def dispatch_on_event_type(self, listener: RealTimeEventListener, event: V4Event
             listener_method_name, payload_field_name = self.dispatch_dict[event.type]
             listener_method = getattr(listener, listener_method_name)
             listener_method(event.initiator, getattr(event.payload, payload_field_name))
+        except AttributeError as e:
+            print(f"Attribute error while dispatching the {event}. {e}")
```
symphony-youri, line 92: should not we start using logging instead of printf?

symphony/bdk/core/service/datafeed/datafeed_loop_v1.py
```diff
@@ -0,0 +1,64 @@
...
+from symphony.bdk.core.service.datafeed.on_disk_datafeed_id_repository import OnDiskDatafeedIdRepository
+from symphony.bdk.gen.exceptions import ApiException
+
+
+class DatafeedLoopV1(AbstractDatafeedLoop):
+    """A class for implementing the datafeed v1 loop service.
+
+    This service will be started by calling :func:`~DatafeedLoopV1.start`.
+
+    On the vert first run, the datafeed will be created and the BDK bot will listen to this datafeed to receive
```
symphony-youri, line 11: vert-> very

symphony/bdk/core/service/datafeed/datafeed_loop_v1.py
```diff
@@ -0,0 +1,64 @@
...
+            self.handle_v4_event_list(events.value)
+
+    async def _create_datafeed_and_persist(self):
+        response = await self.datafeed_api.v4_datafeed_create_post(session_token=await self.auth_session.session_token,
+                                                                   key_manager_token=await self.auth_session.key_manager_token)
+        datafeed_id = response.id
+        self.datafeed_repository.write(response.id)
+        return datafeed_id
+
+    def retrieve_datafeed(self):
```
symphony-youri, line 63: do we need to make it public?

tests/core/config/models/test_bdk_datafeed_config.py
```diff
@@ -0,0 +1,28 @@
+from symphony.bdk.core.config.model.bdk_datafeed_config import BdkDatafeedConfig
+from pathlib import Path
+import pytest
+
+
+@pytest.fixture(params=["v1", 25, True])
```
symphony-youri, line 6: what are 25/true used for?

tests/core/config/models/test_bdk_datafeed_config.py
```diff
@@ -0,0 +1,28 @@
...
+def datafeed_version(request):
+    return request.param
+
+
+# maybe can find an alternative to cucumber
```
symphony-youri, line 11: I didn't understand the comment

tests/resources/datafeed/datafeedId
symphony-youri: Sure? it seems used in the tests

symphony/bdk/core/service/datafeed/on_disk_datafeed_id_repository.py
symphony-elias: @symphony-thibault I understand you don't want to put the interface and implementation in the same module. But I don't like putting code in the `__init__.py` files.
We could split this into two modules and put import statements in the `__init__.py` files to simplify imports.

symphony-youri:
I guess we can close this one @symphony-elias ?

symphony-elias:
Done as part of #114

PR #109, opened 2021-02-01
PLAT-10367: Fixed blast message endpoint

symphony-elias:
Generated api_client.py and rest.py modules were not properly handling text/plain body parts in a multipart/form-data query

### Ticket
PLAT-10367

### Description
Added content type text/plain for string form values.


symphony-youri (approved):
Since we are changing this part, should we add tests for it?
I quickly tried to look for an alternative to wiremock but could not find any

symphony-hong (approved):
Looks good!

PR #110, opened 2021-02-02
PLAT-10466 - Handle Null Events gracefully

anthony-symphony:
### Ticket
PLAT-10466

### Description
In certain cases a null event can be passed to the bot in the Datafeed v5 endpoint and we try to log all event ids where a null will not have one and crash the bot.

### DIFF
--- a/sym_api_client_python/services/abstract_datafeed_event_service.py
+++ b/sym_api_client_python/services/abstract_datafeed_event_service.py
@@ -154,12 +154,15 @@ class AbstractDatafeedEventService(ABC):
         """
         log.debug('DataFeedEventService/handle_events()')
         for event in events:
+            if event is None:
+                continue
+            
             log.debug(
                 'DataFeedEventService/read_datafeed() --> '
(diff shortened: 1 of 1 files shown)


symphony-youri:
@anthony-symphony Could you update your branch against master please? It looks like it is outdated and does not include the circleci configuration.

symphony-youri:
Mmm that's weird I was able to push more commits to your branch but they don't appear in the PR
@anthony-symphony do you see a _allow edits for maintainers_ checkbox in the PR?

anthony-symphony:
This was built off master so whatever is in master should already be there. I originally made a PR against 1.3 branch but, it looks like that was behind.

If you do a diff between master and my commit, it should only be that one file.

Actually I see another commit was made to this PR https://github.com/SymphonyPlatformSolutions/symphony-api-client-python/commit/25c8b9da8ee5d46e16f712f6ae3292d93e29d08d

symphony-youri (approved):

PR #114, opened 2021-02-08
PLAT-10342: DatafeedV1 loop

symphony-elias:
### Ticket
PLAT-10342

### Description
DFv1 loop

### Checklist


symphony-youri (commented):
Will stop() behave correctly with async methods?
(my assumption is that we would still have to wait for the pending request to end to stop the datafeed loop, is there a way we could interrupt the waiting read?)

symphony-youri (approved):

PR #119, opened 2021-02-09
Document message attachment supported parameters

### LINKED ISSUE #120: Forward an attachment without saving it to the disk
How do you retrieve an attachment and forward it to another stream/room without saving its content to the disk?

Closes #120

symphony-youri:
It was not so obvious how the send message API can be used without a
file on the disk but with an in memory stream instead.

### DIFF
--- a/sym_api_client_python/clients/message_client.py
+++ b/sym_api_client_python/clients/message_client.py
@@ -65,15 +65,24 @@ class MessageClient(APIClient):
 
     def send_msg_with_attachment(self, stream_id, msg,
                                  filename, path_to_attachment):
-        """In this function make sure that msg parameter is set to just the messageML string.
+        """
+        In this function make sure that msg parameter is set to just the messageML string.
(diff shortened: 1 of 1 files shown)


thibauult (approved):

sym_api_client_python/clients/message_client.py
```diff
@@ -65,15 +65,24 @@ def _data_and_headers_for_attachment(self, stream_id, msg, filename, path_to_att
 
     def send_msg_with_attachment(self, stream_id, msg,
                                  filename, path_to_attachment):
```
symphony-elias: Maybe we should rename the parameter and add a hint like:
`def send_msg_with_attachment(self, stream_id, msg, filename, attachment: Union[str, io.BytesIO])`

symphony-youri:
Closes #120

symphony-elias (approved):

PR #122, opened 2021-02-11
PLAT-10420: proxy support and SSL verification

symphony-elias:
### Ticket
PLAT-10420

### Description
Proxy support with and without basic authentication.
Enforced SSL verification with a configurable truststore (otherwise it will use the default system CA).

### Checklist

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -61,6 +61,9 @@ pod:
 agent:
    host: dev-agent.symphony.com
    port: 5678
+   proxy:
+     host: agent-proxy
+     port: 3396
 
 keyManager:
   host: dev-key.symphony.com
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -38,6 +38,65 @@ config_3 = BdkConfigLoader.load_from_symphony_dir("config.yaml")  # 3
 ```
 1. Load configuration from a file
 2. Load configuration from a string object
-3. Load configuration from the Symphony directory. The Symphony directory is located under `.symphony` folder in your home directory 
-    . It can be useful when you don't want to share your own Symphony credentials within your project codebase
+3. Load configuration from the Symphony directory. The Symphony directory is located under `.symphony` folder in your
+   home directory. It can be useful when you don't want to share your own Symphony credentials within your project
+   codebase
 
+## Full configuration example
+```yaml
+scheme: https
+host: localhost.symphony.com
+port: 8443
+
+proxy:
+  host: proxy.symphony.com
+  port: 1234
+  username: proxyuser
+  password: proxypassword
+
+pod:
+  host: dev.symphony.com
+  port: 443
+
+agent:
+   host: dev-agent.symphony.com
+   port: 5678
+
+keyManager:
+  host: dev-key.symphony.com
+  port: 8444
+
+sessionAuth:
+  host: dev-session.symphony.com
+  port: 8444
+
+bot:
+  username: bot-name
+  privateKey:
+    path: /path/to/bot/rsa-private-key.pem
+
+app:
+  appId: app-id
+  privateKey:
+    path: path/to/private-key.pem
+```
+
+### Configuration structure
+
+The BDK configuration now includes the following properties:
+- The BDK configuration can contain the global properties for `host`, `port`, `context` and `scheme`.
+These global properties can be used by the client configuration by default or can be overridden if
+user specify the dedicated `host`, `port`, `context`, `scheme` inside the client configuration.
+- `proxy` contains proxy related information. This field is optional.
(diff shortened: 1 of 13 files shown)


symphony/bdk/gen/rest.py
```diff
@@ -49,8 +49,12 @@ def __init__(self, configuration, pools_size=4, maxsize=None):
             maxsize = configuration.connection_pool_maxsize
 
         ssl_context = ssl.create_default_context(purpose=ssl.Purpose.CLIENT_AUTH, cafile=configuration.ssl_ca_cert)
```
symphony-elias, line 51: if `cafile` is None, it uses the system's default CA, see https://docs.python.org/3/library/ssl.html#ssl.create_default_context

symphony/bdk/gen/rest.py
```diff
@@ -49,8 +49,12 @@ def __init__(self, configuration, pools_size=4, maxsize=None):
...
+        ssl_context.verify_mode = ssl.CERT_REQUIRED
+
         if configuration.cert_file:
-            ssl_context.load_verify_locations(configuration.cert_file)
+            ssl_context.load_cert_chain(
```
symphony-elias, line 55: This was to align again to the mustache template (something I mistakenly modified). Not used so far.

symphony/bdk/core/config/model/bdk_ssl_config.py
```diff
@@ -4,17 +4,14 @@
...
     def __init__(self, config):
         """
 
         :param config: the dict containing the SSL specific configuration.
         """
         if config is not None and TRUST_STORE in config:
-            truststore_config = config.get(TRUST_STORE)
-            self.trust_store = BdkCertificateConfig(path=truststore_config.get("path"),
-                                                    password=truststore_config.get("password"))
+            self.trust_store = BdkCertificateConfig(path=config[TRUST_STORE].get("path"))
```
symphony-youri: we don't have a password anymore?

symphony-youri: in the java we also support passing a content (instead of a path) do we want to do it here too? (at least to support the same config files?)

symphony-youri (approved):

docs/configuration.md
```diff
@@ -38,6 +38,70 @@ config_3 = BdkConfigLoader.load_from_symphony_dir("config.yaml")  # 3
...
+  username: proxyuser
+  password: proxypassword
+
+pod:
+  host: dev.symphony.com
+  port: 443
+
+agent:
+   host: dev-agent.symphony.com
+   port: 5678
```
thibauult, line 63: You could also add an example here showing that a specific proxy can be set per component

docs/configuration.md
```diff
@@ -38,6 +38,70 @@ config_3 = BdkConfigLoader.load_from_symphony_dir("config.yaml")  # 3
...
+  host: dev-key.symphony.com
+  port: 8444
+
+sessionAuth:
+  host: dev-session.symphony.com
+  port: 8444
+
+ssl:
+  trustStore:
+    path: /path/to/truststore.pem
```
thibauult, line 78:
```yaml
ssl:
  trustStore:
    path: /path/to/truststore.pem
    password: changeit
```
From https://github.com/SymphonyPlatformSolutions/symphony-api-client-java/blob/master/docs/configuration.md#full-configuration-example

symphony-elias: Indeed but a password is not needed. We directly used a file in pem format for the truststore, see : https://docs.python.org/3/library/ssl.html#ssl.SSLContext.load_verify_locations

symphony/bdk/core/config/model/bdk_ssl_config.py
symphony-elias: No need for password: https://docs.python.org/3/library/ssl.html#ssl.SSLContext.load_verify_locations 
In Java, we don't support passing directly a content for truststores: https://github.com/SymphonyPlatformSolutions/symphony-api-client-java/blob/master/symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkSslConfig.java
I guess you referred to RSA private keys when authenticating.
And the generated Configuration object does not support passing directly the content of the PEM. So I'd prefer to open a ticket if we ever need this.

docs/configuration.md
symphony-youri: ah you are right, in Java we use a keystore and here a PEM file!

symphony-youri: Let's perhaps add this link to the documentation (in a comment next to the full example)? saying we support pem (cafile) parameter?

symphony-elias: Done below in the "Configuration structure" part.

symphony-youri (approved):

PR #123, opened 2021-02-15
PLAT-10460: Implemented OBO-enabled endpoints

symphony-elias:
### Ticket
PLAT-10460

### Description
Implemented OBO-enabled endpoints.

### Checklist

### DIFF
--- a/symphony/bdk/core/auth/auth_session.py
+++ b/symphony/bdk/core/auth/auth_session.py
@@ -97,4 +97,4 @@ class OboAuthSession(AuthSession):
 
     @property
     async def key_manager_token(self):
-        return None
+        return ""
--- a/symphony/bdk/core/auth/auth_session.py
+++ b/symphony/bdk/core/auth/auth_session.py
@@ -1,4 +1,4 @@
-from symphony.bdk.core.auth.exception import AuthInitializationException
+from symphony.bdk.core.auth.exception import AuthInitializationError
 
 
 class AuthSession:
@@ -73,10 +73,10 @@ class OboAuthSession(AuthSession):
         """
         super().__init__(authenticator)
         if user_id is not None and username is not None:
-            raise AuthInitializationException('Username and user id for OBO authentication should not be defined at '
+            raise AuthInitializationError('Username and user id for OBO authentication should not be defined at '
                                               'a same time.')
         if user_id is None and username is None:
-            raise AuthInitializationException('At least username or user id should be defined for '
+            raise AuthInitializationError('At least username or user id should be defined for '
                                               'OBO authentication.')
         self.user_id = user_id
         self.username = username
--- a/symphony/bdk/core/service/obo_services.py
+++ b/symphony/bdk/core/service/obo_services.py
@@ -1,0 +1,69 @@
+from symphony.bdk.core.auth.auth_session import OboAuthSession
+from symphony.bdk.core.client.api_client_factory import ApiClientFactory
+from symphony.bdk.core.config.model.bdk_config import BdkConfig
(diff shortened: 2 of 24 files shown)


symphony/bdk/core/auth/auth_session.py
```diff
@@ -97,4 +97,4 @@ async def session_token(self):
 
     @property
     async def key_manager_token(self):
-        return None
+        return ""
```
symphony-elias, line 100: Having `None` was triggering an error on the generated rest client side.

symphony-youri: As we "control" the generated client, is it something we could handle?

symphony/bdk/core/service/obo_services.py
```diff
@@ -0,0 +1,69 @@
...
+
+class OboServices:
+    """Entry point for OBO-enabled services, see
+    `the list of OBO-enabled endpoints <https://developers.symphony.com/restapi/reference#obo-enabled-endpoints`_
+    """
+
+    async def __aenter__(self):
+        return self
+
+    async def __aexit__(self, exc_type, exc_val, exc_tb):
```
symphony-youri, line 19: how do we document that you should use async with here?

symphony/bdk/core/service/obo_services.py
```diff
@@ -0,0 +1,69 @@
...
+    def __init__(self, config: BdkConfig, obo_session: OboAuthSession):
+        """
+
+        :param config: the BDK configuration.
+        :param obo_session: the OBO session to use.
+        """
+        self._config = config
+        self._obo_session = obo_session
+
+        self._api_client_factory = ApiClientFactory(config)
```
symphony-youri, line 31: should we recreate one or reuse the one from BDK? (I'm wondering why we did it like that for the Java BDK)

symphony/bdk/core/service/obo_services.py
symphony-elias: I added code samples in `docs/authentication.md`.

symphony/bdk/core/service/obo_services.py
symphony-elias: In java, we recreate a new one: 

- https://github.com/SymphonyPlatformSolutions/symphony-api-client-java/blob/master/symphony-bdk-core/src/main/java/com/symphony/bdk/core/SymphonyBdk.java#L258
- https://github.com/SymphonyPlatformSolutions/symphony-api-client-java/blob/364fc4309fb44bab2054d310d8deededc6935c88/symphony-bdk-core/src/main/java/com/symphony/bdk/core/OboServices.java#L31

symphony/bdk/core/auth/auth_session.py
symphony-elias: Error is raised here: https://github.com/SymphonyPlatformSolutions/symphony-api-client-python/blob/2.0/symphony/bdk/gen/model_utils.py#L1276
Since it seems complicated to make the correct checks without breaking anything, I would prefer to keep this as is. The generator in java is a bit different: contrary to the python generated api, the java one checks if the tokens are null before putting them as headers.

symphony-youri: ok

symphony/bdk/core/service/obo_services.py
symphony-youri: Yep that's what I was wondering, do we have a good reason to do it?
If we reuse the existing one from BDK it means we don't have to use async with

symphony-youri (approved):

PR #124, opened 2021-02-15
PLAT-10343: Datafeed V2 service implementation

symphony-hong:
**Draft PR for early review without unittest yet**

### Ticket
[PLAT-10343](https://perzoinc.atlassian.net/browse/PLAT-10343)

### Description
Datafeed V2 service implementation.

In the BdkDatafeedConfig class, check if the `datafeed.version` configuration is
specified. If not, the datafeed version equals `v1`

In the ServiceFactory class, check if the datafeed version is configured as `v2`, return
the DatafeedLoopV2 instance. Otherwise, the DatafeedLoopV1 will be used by default.

Implementation the datafeed v2 loop service in DatafeedLoopV2 class

### Dependencies

### Checklist

### DIFF
--- a/symphony/bdk/core/service/datafeed/datafeed_loop_v2.py
+++ b/symphony/bdk/core/service/datafeed/datafeed_loop_v2.py
@@ -1,3 +1,5 @@
+from typing import Optional
+
 from symphony.bdk.core.auth.auth_session import AuthSession
 from symphony.bdk.core.config.model.bdk_config import BdkConfig
 from symphony.bdk.core.service.datafeed.abstract_datafeed_loop import AbstractDatafeedLoop
@@ -5,7 +7,6 @@ from symphony.bdk.gen import ApiException
 from symphony.bdk.gen.agent_api.datafeed_api import DatafeedApi
 from symphony.bdk.gen.agent_model.ack_id import AckId
 from symphony.bdk.gen.agent_model.v5_datafeed import V5Datafeed
-from typing import Optional
 
 
 class DatafeedLoopV2(AbstractDatafeedLoop):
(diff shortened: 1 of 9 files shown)


symphony/bdk/core/service/datafeed/datafeed_loop_v2.py
```diff
@@ -0,0 +1,90 @@
...
+
+    async def _delete_datafeed(self) -> None:
+        session_token = await self.auth_session.session_token
+        key_manager_token = await self.auth_session.key_manager_token
+        await self.datafeed_api.delete_datafeed(datafeed_id=self.datafeed_id,
+                                                session_token=session_token,
+                                                key_manager_token=key_manager_token)
+
+    async def _recreate_datafeed(self):
+        await self._delete_datafeed()
```
symphony-elias, line 90: Shouldn't we reset `self._ack_id` ?

tests/core/config/models/test_bdk_datafeed_config.py
```diff
@@ -3,7 +3,7 @@
 import pytest
 
 
-@pytest.fixture(params=["v1", 25, True])
+@pytest.fixture(params=["v1"])
```
symphony-elias, line 6: You should test with v2 and check it returns the DatafeedLoopV2

symphony-elias: This was actually to test default value of datafeed version is v1.

symphony-hong: I don't like to have much of logic in the model class. That's why I move the check to the `ServiceFactory` class which will instantiate the DatafeedLoop depending on the version presented in the config object.
https://github.com/SymphonyPlatformSolutions/symphony-api-client-python/pull/124/files#diff-ef818d9162ea0611c768242c2a4089c29eae83aaa77223175f0f78dbb94af5c7R99-R110

symphony-hong: I added unittest for the `ServiceFactory` class.

symphony-elias (approved):

thibauult (commented):
No need to modify https://github.com/SymphonyPlatformSolutions/symphony-api-client-python/blob/2.0/docs/configuration.md?

Just to let appears that DF version can be changed through `config.yaml` file

thibauult (approved):
👌

PR #128, opened 2021-02-17
Addd Snyk monitor command

symphony-elias:
### Ticket
PLAT-10187

### Description
Added monitor command so that we are notified when vulnerabilities are discovered. Documentation here: https://github.com/snyk/actions

### Checklist

### DIFF
--- a/.github/workflows/snyk.yml
+++ b/.github/workflows/snyk.yml
@@ -15,3 +15,4 @@ jobs:
           SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
         with:
           args: --severity-threshold=high
+          command: monitor


symphony-elias:
Previous snyk scan went fine: https://github.com/SymphonyPlatformSolutions/symphony-api-client-python/runs/1917942249?check_suite_focus=true

symphony-soufiane (approved):

symphony-youri:
Do we need it as we are using the GH integration?

symphony-elias:
> Do we need it as we are using the GH integration?

GH integration believes we are using a requirements.txt file, so it skips the scan since it doesn't find any. I think this will be removed when master branch will contain a pyproject.toml file.

symphony-youri (approved):

PR #132, opened 2021-02-22
PLAT-10483 - HealthService implementation

symphony-youness:
### Ticket
[PLAT-10483](https://perzoinc.atlassian.net/browse/PLAT-10483)

### Description
The Healthservice is initialised in `SymphonyBdk._initialize_bot_services`
We can use it by calling `bdk.health()`

Because it's a minor service:
- I didn't make an example file
- Didn't make a specific documentation.md file

### Checklist


symphony-youness:
Here is an example to run in a python, should we add it as an example ?

```python
async def run():

        health_service = bdk.health()

        health_check_resp = await health_service.health_check()
        health_check_extended_resp = await health_service.health_check_extended()
        agent_info_resp = await health_service.get_agent_info()

        logging.debug("\n health_check_resp: %s \n health_check_extended_resp: %s \n agent_info: %s", health_check_resp,
                      health_check_extended_resp, agent_info_resp)
```

tests/core/service/health/health_service_test.py
```diff
@@ -0,0 +1,90 @@
...
+def health_service(mocked_system_api_client, mocked_signals_api_client):
+    return HealthService(mocked_system_api_client, mocked_signals_api_client)
+
+
+@pytest.mark.asyncio
+async def test_health_check(health_service, mocked_system_api_client):
+    mocked_system_api_client.v3_health.return_value = object_from_json_relative_path(
+        "health_response/health_check.json")
+
+    health: V3Health = await health_service.health_check()
```
symphony-youri: is that necessary to put the type here? isn't here inferred from the method's return type?

symphony-youri (approved):
No need for an example or specific doc IMHO

tests/core/service/health/health_service_test.py
symphony-youness: No it's not necessary. I will remove it and all type annotations like these in the file.

tests/core/service/health/health_service_test.py
```diff
@@ -0,0 +1,90 @@
...
+    assert health.services.datafeed.status == "UP"
+    assert health.users.agentservice.status == "UP"
+
+
+@pytest.mark.asyncio
+async def test_health_check_extended_failed(health_service, mocked_system_api_client):
+    mocked_system_api_client.v3_extended_health.side_effect = ApiException(400)
+
+    with pytest.raises(ApiException):
+        await health_service.health_check_extended()
```
thibauult, line 68: A question that have been asked on Java BDK side: do we really want to return an exception when HealthCheck is down? Isn't something that we want to handle at service level and just return `health.status == "DOWN"`?

symphony-youness: current behavior => raises an exception if the status is "DOWN"

From a Developer experience POV, as per our discussion:
**we will wait for users' feedback to change the current behavior to return "DOWN" instead of raising an exception.**

As far as I am concerned:
- If the agent is DOWN, we can't have a pod status.
- Even if the agent is DOWN we can still perform calls to the POD for other actions.
From a functional POV knowing that the agent is down and the pod is not is important (maybe that should feature request)

symphony/bdk/core/service/health/health_service.py
```diff
@@ -0,0 +1,40 @@
...
+        `Health Check v3 <https://developers.symphony.com/restapi/reference#health-check-v3>`_
+        """
+        return await self._system_api.v3_health()
+
+    async def health_check_extended(self) -> V3Health:
+        """Returns the connectivity status of the Agent services as well as users connectivity.
+        Available on Agent 2.57.0 and above.
+
+        :return: V3Health: the connectivity status of the Agent services as well as users connectivity.
+        `Healt Check Extended v3 <https://developers.symphony.com/restapi/reference#health-check-extended-v3>`_
```
symphony-elias: It is better to put the links to the documentation before the `:return:` statement IMHO

symphony-elias (approved):
LGTM

PR #137, opened 2021-03-01
PLAT-10529 Ability to set private key and certificate content programmatically

symphony-soufiane:
### Ticket
[PLAT-10529](https://perzoinc.atlassian.net/browse/PLAT-10529)

### Description
Private key and Certificate were able to only be loaded from the config file when initializing bdk. In this PR, we update bot config to be able to set private key and certificate after bdk being initialized.
BdkAuthenticationConfig constructor overrides both private key and certificate even when only one is provided, the missing one is overriden to the default value and thus its content is lost. This is why two setters have been added to the class to be able to set one independantly of the other.

### Dependencies

### Checklist


examples/auth.py
```diff
@@ -15,6 +15,18 @@ async def run():
         obo_auth_session = bdk.obo(username="username")
         logging.info(await obo_auth_session.session_token)
 
+        # update private key
+        config.setBotConfig(privateKey="/path/to/new_private_key")
```
thibauult: @symphony-soufiane I don't really understand how we can set the actual content of the key here? 

Something like: 
```python
rsa_private_key = "-----BEGIN RSA PRIVATE KEY-----...-----END RSA PRIVATE KEY-----"
config.setBotConfig(privateKeyContent=rsa_private_key)
```

symphony-soufiane: I'm not sure if we want to put the private key content in the code ? Here we give the path to the pem file like :
`config.setBotConfig(privateKey="/Users/soufiane/.symphony/rsa-localhost/privatekey.pem"`

Should it be changed to take privateKey/certificate as a string as you say ?

tests/core/config/bdk_config_test.py
```diff
@@ -0,0 +1,32 @@
+from symphony.bdk.core.config.loader import BdkConfigLoader
+
+from tests.utils.resource_utils import get_config_resource_filepath
+import pytest
+
```
symphony-soufiane, line 6: I was missing inspiration for unit test as at the end I'm testing setters..

examples/auth.py
thibauult: Actually, the goal of this feature is to give the ability to developers to load either RSA private key or certificate content from an external system (for instance [Vault](https://www.vaultproject.io/) or [AWS Secrets Manager](https://aws.amazon.com/secrets-manager/)). Without having to copy them on the file system as they are quite sensitive data.

symphony/bdk/core/config/model/bdk_config.py
```diff
@@ -26,6 +26,12 @@ def __init__(self, **config):
         self.app = BdkAppConfig(config.get("app"))
         self.datafeed = BdkDatafeedConfig(config.get("datafeed"))
 
+    def setBotConfig(self, private_key_content=None, certificate_content=None):
+        if private_key_content is not  None :
+            self.bot.setPrivatekey(path=None, private_key_content=private_key_content)
```
symphony-soufiane, line 31: do we agree that the path will always be None as the content is given by string ?

thibauult (commented):
Do you where we decide whether we use the `path` or the `content`? For instance in the Java BDK we have this [section of code](https://github.com/finos/symphony-bdk-java/blob/main/symphony-bdk-core/src/main/java/com/symphony/bdk/core/auth/AuthenticatorFactory.java#L169-L174) that select the `content` prior to `path` if both are configured

examples/auth.py
```diff
@@ -15,6 +15,30 @@ async def run():
         obo_auth_session = bdk.obo(username="username")
         logging.info(await obo_auth_session.session_token)
 
+        # update private key and certificate
+        private_key_string = '-----BEGIN RSA PRIVATE KEY-----\n'\
```
symphony-elias: As of today we don't support support certificate authentication. So please remove the certificate examples from the example.

tests/core/config/bdk_config_test.py
symphony-elias: The configuration should be either RSA or certificate exclusively (we cannot have both). And you should also test that BotAuthenticatorRsa uses the private key content when specified while authenticating

symphony-elias:
> Do you where we decide whether we use the `path` or the `content`? For instance in the Java BDK we have this [section of code](https://github.com/finos/symphony-bdk-java/blob/main/symphony-bdk-core/src/main/java/com/symphony/bdk/core/auth/AuthenticatorFactory.java#L169-L174) that select the `content` prior to `path` if both are configured

It is defined in `jwt_helper.create_signed_jwt`. Maube we could move that logic into the `BdkRsaKeyConfig`

symphony-elias:
Please update docsrc/markdown/authentication.md

symphony-soufiane:
> > Do you where we decide whether we use the `path` or the `content`? For instance in the Java BDK we have this [section of code](https://github.com/finos/symphony-bdk-java/blob/main/symphony-bdk-core/src/main/java/com/symphony/bdk/core/auth/AuthenticatorFactory.java#L169-L174) that select the `content` prior to `path` if both are configured
> 
> It is defined in `jwt_helper.create_signed_jwt`. Maube we could move that logic into the `BdkRsaKeyConfig`

As far as I understood, we cannot have both path and content at the same time, this is considered as an invalid configuration ( see link below). Thus we will not have to decide which one to use. And that's why we put path=None when providing rsa pk content.
https://github.com/SymphonyPlatformSolutions/symphony-api-client-python/blob/4d614bf972d13784ad4c1f96b8158852ffcaf6c0/symphony/bdk/core/config/model/bdk_authentication_config.py#L26

docsrc/markdown/authentication.md
```diff
@@ -38,6 +38,33 @@ async def run():
...
+from symphony.bdk.core.symphony_bdk import SymphonyBdk
+
+async def run():
+    # loading configuration
+    config = BdkConfigLoader.load_from_symphony_dir("config.yaml")
+    async with SymphonyBdk(config) as bdk:
+        # update private key with content
+        private_key_string = '-----BEGIN RSA PRIVATE KEY-----\n'\
+                             'zI2OZtdb8fu/xl7itIAOzKLFg3mhA...\n'\
+        config.setBotConfig(private_key_content=private_key_string)
```
symphony-youri: is the setBotConfig pythonic?
Should it be set_bot_config?

do we want to provide such? should not it be rather an example with a config file?

symphony-youri (commented):
why don't we 

config.bot.private_key.content=... ?

docsrc/markdown/authentication.md
symphony-soufiane: You're right, I put it to set_bot_config. 
Do you mean that I need to create an example for this and keep this one for config file usage ?

symphony-soufiane:
> why don't we
> 
> config.bot.private_key.content=... ?

Because I wanted to keep the logic of setting path to None when we put content. In this case, it would be done in 2 lines : 
config.bot.private_key.content=...
config.bot.private_key.path=None

thibauult:
> > why don't we
> > config.bot.private_key.content=... ?
> 
> Because I wanted to keep the logic of setting path to None when we put content. In this case, it would be done in 2 lines :
> config.bot.private_key.content=...
> config.bot.private_key.path=None

@symphony-soufiane let's follow the [Java BDK](https://github.com/finos/symphony-bdk-java/blob/main/symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkRsaKeyConfig.java#L16-L17)

docsrc/markdown/authentication.md
```diff
@@ -38,6 +38,33 @@ async def run():
...
+from symphony.bdk.core.symphony_bdk import SymphonyBdk
+
+async def run():
+    # loading configuration
+    config = BdkConfigLoader.load_from_symphony_dir("config.yaml")
+    async with SymphonyBdk(config) as bdk:
+        # update private key with content
+        private_key_string = '-----BEGIN RSA PRIVATE KEY-----\n'\
+                             'zI2OZtdb8fu/xl7itIAOzKLFg3mhA...\n'\
+        config.bot.private_key.setContent((rsa_key_content=private_key_string)
```
symphony-elias: this should be put before the first `async with SymphonyBdk(config)` statement. You can remove the second `async with` statement.

symphony/bdk/core/config/model/bdk_certificate_config.py
```diff
@@ -21,3 +21,20 @@ def is_valid(self) -> bool:
         :return: true if the certificate configuration is valid.
         """
         return not (self.path is not None and self.content != "")
+
+    def setContent(self, certificate_content):
+        """Sets ceritficate content and overrides path to None to keep certificate config valid
+        Args:
```
symphony-elias: Please use restructured text format. You can configure it on PyCharm for instance: https://www.jetbrains.com/help/pycharm/settings-tools-python-integrated-tools.html -> docstring format.

tests/core/auth/bot_authenticator_test.py
```diff
@@ -82,3 +79,27 @@ async def test_authenticate_bot(config, mocked_api_client):
 
         assert await auth_session.session_token == "session_token"
         assert await auth_session.key_manager_token == "km_token"
+
+@pytest.mark.asyncio
+async def test_authenticate_with_private_key_content(config, mocked_api_client):
+    with patch('symphony.bdk.core.auth.bot_authenticator.create_signed_jwt', return_value='privateKey'):
```
symphony-elias, line 86: The logic inside `create_signed_jwt` to fetch the private key content should be put in the `BdkRsaKeyConfig` class with a specific getter so that you could test create_signed_jwt is called with the correct value.

symphony/bdk/core/config/model/bdk_certificate_config.py
```diff
@@ -21,3 +21,20 @@ def is_valid(self) -> bool:
         :return: true if the certificate configuration is valid.
         """
         return not (self.path is not None and self.content != "")
+
+    def setContent(self, certificate_content):
```
symphony-elias: Please follow python conventions: https://www.python.org/dev/peps/pep-0008/#naming-conventions
Function names should be lowercase with underscores.

symphony/bdk/core/config/model/bdk_certificate_config.py
symphony-youri: typo in certificate too

symphony/bdk/core/config/model/bdk_certificate_config.py
symphony-youri: Cannot we just use override the setter for this? (to make sure we have a content or a path but not both)

symphony-soufiane:
> should not we use Python properties? (https://docs.python.org/3/library/functions.html#property)

I didn't know about this. But how can we use it for the _get_private_key_from_config_ ? It either returns content from path or the content that was set, for me it is not a getter but an instance method. Maybe we can use properties only for setters? @symphony-youri

symphony/bdk/core/config/model/bdk_certificate_config.py
```diff
@@ -21,3 +21,20 @@ def is_valid(self) -> bool:
         :return: true if the certificate configuration is valid.
         """
         return not (self.path is not None and self.content != "")
+
+    def set_content(self, certificate_content):
+        """Sets ceritficate content and overrides path to None to keep certificate config valid
+        Args:
```
symphony-youness: The markup used for the python BDK documentation is reStructuredText. 
Here is my [cheat sheet](https://thomas-cokelaer.info/tutorials/sphinx/rest_syntax.html), could be helpful. You can also check other file docstrings and it should be fine.
This means you should use:
```python
def function(param1):
''' Some description

:param param1 description of parameter
'''
```

symphony/bdk/core/config/model/bdk_rsa_key_config.py
```diff
@@ -19,3 +19,36 @@ def is_valid(self) -> bool:
...
+        Args:
+            rsa_key_path: rsa private key path
+        """
+        self.path = rsa_key_path
+        self.content = None
+
+    def get_private_key_from_config(self):
+        """Loads the private key content.
+        If the path is set, it loads the file content, otherwise it returns the content.
+        Returns: private key content as string.
```
symphony-youness: I saw you used the `:returns` (reStructuredText) everywhere 👌 
can you please add a lowercase R here

tests/core/config/models/bdk_authentication_config_test.py
```diff
@@ -10,18 +10,16 @@ def test_rsa_configuration():
     # Set up a non_empty content
     authentication_config.private_key.content = "not_empty"
     assert authentication_config.is_rsa_authentication_configured() is True
-    assert authentication_config.is_rsa_configuration_valid() is False
```
symphony-soufiane, line 13: I removed this statement as setting the content overrides the path, thus config is still valid

symphony/bdk/core/config/model/bdk_rsa_key_config.py
```diff
@@ -2,20 +2,55 @@ class BdkRsaKeyConfig:
...
+        """Loads the private key content.
+        If the path is set, it loads the file content, otherwise it returns the content.
+
+        :return: private key content as string.
+        """
+        return self._load_key_from_path() \
+            if self._path is not None else self._content
+
+    def _load_key_from_path(self):
+        if self._path is not None:
```
symphony-elias: This check is already done before calling the method so you can safely remove it

symphony/bdk/core/config/model/bdk_rsa_key_config.py
```diff
@@ -2,20 +2,55 @@ class BdkRsaKeyConfig:
...
+        self._path = None
+
+    def set_path(self, rsa_key_path):
+        """Sets rsa path and overrides content to None to keep key's config valid
+        :param rsa_key_path: rsa private key path
+        """
+        self._path = rsa_key_path
+        self._content = None
+
+    def get_private_key_from_config(self):
```
symphony-elias: I would prefer `get_private_key_content`

symphony/bdk/core/config/model/bdk_rsa_key_config.py
```diff
@@ -2,20 +2,55 @@ class BdkRsaKeyConfig:
...
+
+    def _load_key_from_path(self):
+        if self._path is not None:
+            with open(self._path, "r") as f:
+                private_key_content = f.readlines()
+                key = "".join(private_key_content)
+                return key
+        return ""
+
+    path = property(fset=set_path, doc="path to rsa private key file")
```
symphony-elias: I don't really like having properties defined this way. Most pythonic way for properties is to use decorators like here: https://docs.python.org/3/library/functions.html#property
It is ok for me to have getters for path and content.

tests/core/auth/bot_authenticator_test.py
```diff
@@ -21,7 +21,6 @@ def config():
     }
     return BdkBotConfig(bot_config)
 
-
 @pytest.fixture()
 def mocked_api_client():
     def __loader():
```
symphony-youness, line 26: I am sorry but I don't understand the use of __loader() here

tests/core/config/models/bdk_authentication_config_test.py
```diff
@@ -45,7 +45,7 @@ def test_certificate_configuration():
     assert authentication_config.is_certificate_configuration_valid() is False
 
     # No password with certificate
-    authentication_config.certificate.path = None
+    authentication_config.certificate.path =None
```
symphony-youness: supper picky, sorry: please add a space after the equal sign. 
you can (shift + option + command + L) on pycharm to reformat the file and you will never have these again 😄

symphony-youness (approved):

symphony-elias (approved):
LGTM !

PR #139, opened 2021-03-04
PLAT-10489: Implemented automatic pagination

symphony-elias:
### Ticket
PLAT-10489

### Description
Implemented automatic pagination for paginated endpoints.

### Checklist


tests/core/service/stream/stream_service_test.py
```diff
@@ -203,6 +226,22 @@ async def test_search_rooms(mocked_api_client, stream_service, streams_api):
...
 
 
+@pytest.mark.asyncio
+async def test_search_all_rooms(mocked_api_client, stream_service, streams_api):
+    mocked_api_client.call_api.return_value = object_from_json_relative_path("stream/search_rooms.json")
+
+    search_criteria = V2RoomSearchCriteria(query="query")
+    chunk_size = 3
+
+    gen = await stream_service.search_all_rooms(search_criteria, chunk_size=chunk_size)
```
symphony-youri, line 236: maybe we should add an example in the doc for this? (not entirely sure because at the same time using this method  is potentially costly in term of API calls and we might not want to promote using it too much)

symphony-youri (approved):
👍

PR #140, opened 2021-03-05
Upgraded aiohttp to 3.7.4 to fix dependabot issue

symphony-elias:
Manually tested with async DF loop. DF loop + send message works fine locally.

### Ticket
https://github.com/SymphonyPlatformSolutions/symphony-api-client-python/security/dependabot/requirements.txt/aiohttp/open

### Description
Upgraded aiohttp to 3.7.4 to fix dependabot issue

### Checklist

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -13,7 +13,7 @@ ecdsa<0.15
 multidict~=4.7.6
 yarl~=1.4.2
 attrs~=19.3.0
-aiohttp~=3.6.2
+aiohttp>=3.7.4
 chardet~=3.0.4
 MarkupSafe~=1.1.1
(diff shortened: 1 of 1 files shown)


symphony-youri:
Would that be worth having a few tests using something like https://github.com/gabrielfalcao/HTTPretty (I just google for 'mock http server python')?
So that we can automate this kind of testing? (would be interesting if we can configure custom certs)
I might take a shot at it while fixing the CA system certs loading

symphony-youri (approved):
(just discovered https://python-poetry.org/docs/cli/#update that could help us maintain the dependencies up to date)

PR #141, opened 2021-03-05
PLAT-10533: Added User-Agent header in requests

symphony-elias:
### Ticket
PLAT-10533

### Description
Added user-agent header with BDK version and python version.

### Checklist

### DIFF
--- a/symphony/bdk/core/client/api_client_factory.py
+++ b/symphony/bdk/core/client/api_client_factory.py
@@ -19,7 +19,7 @@ class ApiClientFactory:
         self._login_client = self._get_api_client(self._config.pod, '/login')
         self._pod_client = self._get_api_client(self._config.pod, '/pod')
         self._relay_client = self._get_api_client(self._config.key_manager, '/relay')
-        self._agent_client = self._get_api_client(self._config.session_auth, '/agent')
+        self._agent_client = self._get_api_client(self._config.agent, '/agent')
         self._session_auth_client = self._get_api_client(self._config.session_auth, '/sessionauth')
 
     def get_login_client(self) -> ApiClient:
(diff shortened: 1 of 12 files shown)


symphony-youri:
Please provide a description in the commit with at least an example of the header (saying why we set it would be nice too)

symphony/bdk/core/client/api_client_factory.py
```diff
@@ -71,11 +74,34 @@ def _get_api_client(self, server_config, context) -> ApiClient:
...
         if proxy_config.are_credentials_defined():
-            configuration.proxy_headers = urllib3.util.make_headers(proxy_basic_auth=proxy_config.get_credentials())
+            configuration.proxy_headers = urllib3.util.make_headers(proxy_basic_auth=proxy_config.get_credentials(),
+                                                                    user_agent=ApiClientFactory._user_agent())
+        else:
+            configuration.proxy_headers = urllib3.util.make_headers(user_agent=ApiClientFactory._user_agent())
+
+    @staticmethod
+    def _user_agent():
+        return f"Symphony-BDK-Python/{ApiClientFactory._bdk_version()} Python/{ApiClientFactory._python_version()}"
```
symphony-youri, line 122: The format looks a bit different from the Java one:

return "Symphony BDK/" + getBdkVersion() + "/java/" + System.getProperty("java.version");

symphony-elias: Yes but I realized we should not have spaces : https://tools.ietf.org/html/rfc7231#section-5.5.3
I was thinking this was more in line with the RFD. WDYT ? 
Either we keep like the java one or we change the user-agent in the java BDK

symphony-youri: I agree, let's update it for the Java BDK then (and consider it a non breaking change...)

symphony-elias: GH issue opened here on the Java BDK: https://github.com/finos/symphony-bdk-java/issues/471

symphony-youri (approved):

PR #142, opened 2021-03-08
PLAT-10643: Load certs from system store

symphony-youri:
### Ticket
[PLAT-10643](link to ticket)

### Description
In #122 we enable loading a custom cert store for the HTTP client. This
broked using pods with valid certs because the system certs are no
longer loaded. This change loads them even all the time (and we might
add custom ones on top of that).

Also sync the rest.mustache template used for code generation.

### Dependencies
N/A

### Checklist

### DIFF
--- a/symphony/bdk/gen/rest.py
+++ b/symphony/bdk/gen/rest.py
@@ -49,6 +49,7 @@ class RESTClientObject(object):
             maxsize = configuration.connection_pool_maxsize
 
         ssl_context = ssl.create_default_context(purpose=ssl.Purpose.CLIENT_AUTH, cafile=configuration.ssl_ca_cert)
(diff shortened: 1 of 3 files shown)


symphony-elias (approved):
LGTM

symphony-youri:
For unit testing the change I took a look at https://httpretty.readthedocs.io/en/latest/ but it does not seem appropriate as it is mocking the low level TCP layer
An alternative could be https://stackoverflow.com/questions/19705785/python-3-simple-https-server
but in this case we want to test the system certs so. I added a test checking pinging Google
This a bit controversial because we don't like external dependencies for tests but here maybe it is ok

symphony-elias (approved):
👍

PR #145, opened 2021-03-10
PLAT-10600 enforce pylint checks in PR builder

symphony-mariacristina:
### Ticket
[PLAT-10600](https://perzoinc.atlassian.net/browse/PLAT-10600)

### Description
Goal of this PR is to enable pylint checks in the PR builder. I used a Github action Super-Linter which runs the pylint scan only on the diff found in the PR. For now the limit to fail is set to 9.00 (max being 10.00). We can start like this and if needed we can adapt it later on if we think that is too strict.

**!!!** This is a **test PR** just to show how the Super Linter works. Before merging I will remove the errors I introduced in the bot_authenticator.py to show how it works.

**TODO**
- [ ] check existing errors if we can fix some of them

### Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -45,9 +45,6 @@ jobs:
       - name: Run tests with coverage
         run: poetry run pytest
 
-      - name: Run pylint
-        run: poetry run pylint symphony tests
-
       - name: Upload test results
         uses: actions/upload-artifact@v2
         with:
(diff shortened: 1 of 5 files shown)


symphony-youri:
Do we have to use the custom action to run it on the diff? As it is a recent project, isn't it easier to make it pass right now for all existing sources and keep enforcing it?

PR #146, opened 2021-03-10
PLAT-10600 enforce pylint checks in PR builder

symphony-mariacristina:
### Ticket
[PLAT-10600](https://perzoinc.atlassian.net/browse/PLAT-10600)

### Description
Goal of this PR is to enable pylint checks in the PR builder.
For now the limit to fail is set to 9.50 (max being 10.00). We can start like this and if needed we can adapt it later on, if we think that is too strict.

All changes about fixture are related to this issue: https://stackoverflow.com/questions/46089480/pytest-fixtures-redefining-name-from-outer-scope-pylint

### Checklist

### DIFF
--- a/.github/workflows/pylint.yaml
+++ b/.github/workflows/pylint.yaml
@@ -12,14 +12,30 @@ jobs:
     steps:
       - name: Checkout Code
         uses: actions/checkout@v2
+
+      - name: Set up Python 3.8
+        uses: actions/setup-python@v2
         with:
-          fetch-depth: 0
+          python-version: 3.8
+
+      - name: Install Pylint
+        run: pip install pylint
+
+      - name: Download previous run results
+        uses: dawidd6/action-download-artifact@v2.12.0
+        continue-on-error: true
+        with:
+          workflow: pylint.yaml
+          workflow_conclusion: success
+          name: .pylint.d
+          path: /home/runner/.pylint.d
 
       - name: Run Pylint
-        uses: github/super-linter@v3
-        env:
-          VALIDATE_ALL_CODEBASE: false
-          VALIDATE_PYTHON_PYLINT: true
-          DEFAULT_BRANCH: "2.0"
-          LINTER_RULES_PATH: /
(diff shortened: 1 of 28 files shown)


.github/workflows/pylint.yaml
```diff
@@ -0,0 +1,41 @@
...
+
+      - name: Set up Python 3.8
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+
+      - name: Install Pylint
+        run: pip install pylint
+
+      - name: Download previous run results
```
symphony-mariacristina, line 26: Every run uploads the scan results that can be downloaded in the next run to have a comparison within runs (taking the last successful one).
Generated a message like this at the end of the scan:

```
------------------------------------------------------------------
Your code has been rated at 9.58/10 (previous run: 9.74/10, -0.16)
```

thibauult:
> For now the limit to fail is set to 9.50 

😍

.github/workflows/pylint.yaml
symphony-youri: is it going to comparer the results of the last run for the same branch/PR or for any of the PRs? (the latter would not make much sense right?)

symphony-elias: Yes we have to download .pylint.d of the latest commit on 2.0 branch

.pylintrc
```diff
@@ -478,7 +479,9 @@ variable-naming-style=snake_case
 
 # This flag controls whether inconsistent-quotes generates a warning when the
 # character used as a quote delimiter is used inconsistently within a module.
-check-quote-consistency=no
+check-quote-consistency=yes
```
symphony-elias, line 482: 👍

.github/workflows/pylint.yaml
symphony-mariacristina: yes thank you for noticing, I completely missed that. I will add modify it in the next commit

thibauult (approved):

symphony-elias (approved):
LGTM

symphony-youri (approved):

PR #147, opened 2021-03-10
PLAT-10534: Added X-Trace-Id header to each HTTP call

symphony-elias:
### Ticket
PLAT-10534

### Description
Added X-Trace-Id header to each HTTP calls and implemented a logging filter.

### Checklist

### DIFF
--- a/docsrc/markdown/tech/production_readiness.md
+++ b/docsrc/markdown/tech/production_readiness.md
@@ -19,7 +19,7 @@ to avoid filling the disk.
 ### X-Trace-Id
 The Symphony BDK adds a `trace_id` field to each log record with a value called `X-Trace-Id`
 (random alphanumeric string of 6 characters).
-You are free to print it in your logs using pattern `%(trace-id)s` using a suitable logformat string, for instance like
+You are free to print it in your logs using pattern `%(trace_id)s` using a suitable logformat string, for instance like
 done in the
 [sample logging configuration](https://github.com/SymphonyPlatformSolutions/symphony-api-client-python/blob/2.0/examples/logging.conf).
 
--- a/docsrc/markdown/tech/production_readiness.md
+++ b/docsrc/markdown/tech/production_readiness.md
@@ -15,3 +15,16 @@ a [rotating file handler](https://docs.python.org/3/library/logging.handlers.htm
 to avoid filling the disk.
 
 **Beware of logging sensitive or personal data such as message content, user details,...**
+
+### X-Trace-Id
+The Symphony BDK adds a `trace_id` field to each log record with a value called `X-Trace-Id`
+(random alphanumeric string of 6 characters).
+You are free to print it in your logs using pattern `%(trace-id)s` using a suitable logformat string, for instance like
+done in the
+[sample logging configuration](https://github.com/SymphonyPlatformSolutions/symphony-api-client-python/blob/2.0/examples/logging.conf).
+
+This value is send as header of every request made to the Symphony API. This is especially useful for cross-applications
+debugging, assuming that the `X-Trace-Id` value is also present in your application logs. A new value is generated for
+each request made to the Symphony API, except if you decide to set a custom value using
+[DistributedTracingContext](../../_autosummary/symphony.bdk.core.client.trace_id.DistributedTracingContext).
+In this case, you will have to manage the `X-Trace-Id` yourself and no new value will be generated.
--- a/examples/auth.py
+++ b/examples/auth.py
@@ -1,5 +1,5 @@
 import asyncio
-import logging
+import logging.config
 
 from symphony.bdk.core.config.loader import BdkConfigLoader
 from symphony.bdk.core.symphony_bdk import SymphonyBdk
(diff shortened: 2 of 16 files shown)


symphony-elias:
When you run the auth example with logs between calls, you have this kind of output:
```
2021-03-10 18:23:44,380 - root - INFO - no-trace-id - beginning
2021-03-10 18:23:45,206 - root - INFO - TND2t6 - 0100e26c14ea...
2021-03-10 18:23:45,206 - root - INFO - TND2t6 - After first call
2021-03-10 18:23:45,824 - root - INFO - 6C2Rjx - eyJhbGciO...
2021-03-10 18:23:45,824 - root - INFO - 6C2Rjx - end
```

examples/auth.py
```diff
@@ -16,5 +16,31 @@ async def run():
         logging.info(await obo_auth_session.session_token)
 
 
-logging.basicConfig(level=logging.DEBUG)
+logging.config.dictConfig({
+    "version": 1,
+    "disable_existing_loggers": True,
+    "filters": {"taskfilter": {"()": "symphony.bdk.core.client.api_client_factory.TraceIdFilter"}},
```
symphony-youri: we have to register it here, can we do it programatically to avoid this to the user?

symphony/bdk/core/client/api_client_factory.py
```diff
@@ -9,6 +13,25 @@
 from symphony.bdk.gen.api_client import ApiClient
 from symphony.bdk.gen.configuration import Configuration
 
+trace_id_context = ContextVar("trace_id_context")
+
+
+def add_x_trace_id(func):
```
symphony-youri: do we handle the case where it might be set already?

symphony/bdk/core/client/api_client_factory.py
```diff
@@ -9,6 +13,25 @@
 from symphony.bdk.gen.api_client import ApiClient
 from symphony.bdk.gen.configuration import Configuration
 
+trace_id_context = ContextVar("trace_id_context")
```
symphony-youri: I would perhaps distinguish both, the trace id context that can be used everywhere in the bot, printed in the log 
and the logic to retrieve the trace id from the context, set it as a header to propagate it downstream

thibauult: Is this `trace_id_context` available from anywhere in the code? 

As a developer, I'd potentially like to init value from an HTTP filter if my app is a web application

symphony/bdk/core/client/trace_id.py
```diff
@@ -0,0 +1,86 @@
...
+
+    :param func: the function to be decorated
+    :return: the decorated function
+    """
+
+    async def add_x_trace_id_header(*args, **kwargs):
+        if not DistributedTracingContext.has_trace_id():
+            DistributedTracingContext.set_trace_id()
+
+        args[4]["X-Trace-Id"] = DistributedTracingContext.get_trace_id()
```
symphony-youri: is the args[4] risky to use? (should we check it is here for instance?)

symphony/bdk/core/client/__init__.py
```diff
@@ -1,2 +1,5 @@
 """Package for instantiating internal API clients
 """
+from symphony.bdk.core.client.trace_id import setup_trace_id_log_record_factory
+
+setup_trace_id_log_record_factory()
```
symphony-youri, line 5: just for my knowledge when will this be called? whenever a member of this package is loaded?

symphony/bdk/core/client/api_client_factory.py
```diff
@@ -76,11 +77,17 @@ def _get_api_client(self, server_config, context) -> ApiClient:
...
+        return client
+
+    @staticmethod
+    def _add_headers( client, server_config):
         for header_name, header_value in ApiClientFactory._sanitized_default_headers(server_config).items():
             client.set_default_header(header_name, header_value)
         client.user_agent = ApiClientFactory._user_agent(server_config)
 
-        return client
+        client._ApiClient__call_api = add_x_trace_id(client._ApiClient__call_api)
```
symphony-youri: should we check that the X-Trace-Id is not already set by the lines above (where you can customize header) before adding our own?
just in case a customer is hardcoding a X-Trace-Id using the config

symphony/bdk/core/client/trace_id.py
```diff
@@ -0,0 +1,86 @@
...
+        return await func(*args, **kwargs)
+
+    return add_x_trace_id_header
+
+
+class DistributedTracingContext:
```
symphony-youri, line 51: you have plans to add unit tests for it right? :)

symphony-youri:
Just added a few questions but it looks good to me

symphony/bdk/core/client/__init__.py
symphony-elias: Yes, `__init__.py` is run first time we import this package or any child package.

symphony/bdk/core/client/trace_id.py
symphony-youness: +1 It doesn't seem readble to me. 
I am curious to know how are we planning to use it.
I understand you want to use `add_x_trace_id` as a decorator, how do arguments make sense here ?

docsrc/markdown/tech/production_readiness.md
```diff
@@ -15,3 +15,16 @@ a [rotating file handler](https://docs.python.org/3/library/logging.handlers.htm
 to avoid filling the disk.
 
 **Beware of logging sensitive or personal data such as message content, user details,...**
+
+### X-Trace-Id
+The Symphony BDK adds a `trace_id` field to each log record with a value called `X-Trace-Id`
+(random alphanumeric string of 6 characters).
+You are free to print it in your logs using pattern `%(trace-id)s` using a suitable logformat string, for instance like
```
symphony-youri: trace_id

symphony-youri (approved):
Just a small typo in the documentation

PR #150, opened 2021-03-16
PLAT-10565 Adding utils classes

symphony-mariacristina:
### Ticket
[PLAT-10565](https://perzoinc.atlassian.net/browse/PLAT-10565)

### Description
Adding some util classes on message processing to:
- Extract entities from a given incoming message (mentions, hashtags, cashtags, emojis)
- Extract message content from presentationML
- Pre-process an outgoing message by "cleaning" the text message, escaping all special characters that will violate messageML format
- Convert a stream id into a url safe id and viceversa

### Checklist

### DIFF
--- a/symphony/bdk/core/auth/exception.py
+++ b/symphony/bdk/core/auth/exception.py
@@ -7,6 +7,7 @@ class AuthInitializationError(Exception):
     """
 
     def __init__(self, message: str):
+        Exception.__init__(self, message)
         self.message = message
 
 
@@ -15,5 +16,6 @@ class AuthUnauthorizedError(Exception):
     """
 
     def __init__(self, message: str, cause: Exception = None):
+        Exception.__init__(self, message, cause)
         self.message = message
         self.cause = cause
--- a/symphony/bdk/core/auth/exception.py
+++ b/symphony/bdk/core/auth/exception.py
@@ -7,7 +7,6 @@ class AuthInitializationError(Exception):
     """
 
     def __init__(self, message: str):
-        Exception.__init__(self, message)
         self.message = message
 
 
@@ -16,6 +15,5 @@ class AuthUnauthorizedError(Exception):
     """
 
     def __init__(self, message: str, cause: Exception = None):
-        Exception.__init__(self, message, cause)
         self.message = message
         self.cause = cause
--- a/symphony/bdk/core/service/message/message_parser.py
+++ b/symphony/bdk/core/service/message/message_parser.py
@@ -21,12 +21,11 @@ def get_text_content_from_message(message: V4Message) -> str:
     :return: the message text content extracted from the given PresentationML
     """
 
-    unparsable_presentationml = "Unable to parse the PresentationML, it is not in the correct format."
     presentation_ml = message.message
     try:
         tree = ElementTree(fromstring(presentation_ml))
     except ET.ParseError:
-        raise MessageParserError(unparsable_presentationml)
+        raise MessageParserError("Unable to parse the PresentationML, it is not in the correct format.")
     return tostring(tree.getroot(), method="text").decode().strip()
 
 
@@ -90,11 +89,10 @@ def _get_tags(message, entity_type):
 
 
 def _parse_json_data(json_data):
-    unparsable_json = "Failed to extract payload from message data"
     try:
         return json.loads(json_data)
     except JSONDecodeError:
-        raise MessageParserError(unparsable_json)
+        raise MessageParserError("Failed to extract payload from message data")
 
 
 class _EntityTypeEnum(Enum):
--- a/symphony/bdk/core/service/message/message_parser.py
+++ b/symphony/bdk/core/service/message/message_parser.py
@@ -6,6 +6,7 @@ import json
 
 from enum import Enum
 from json import JSONDecodeError
+from typing import Dict
 from xml.etree import ElementTree as ET
 from xml.etree.ElementTree import ElementTree, fromstring, tostring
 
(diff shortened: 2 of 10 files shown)


symphony/bdk/core/service/message/message_parser.py
```diff
@@ -0,0 +1,103 @@
...
+from enum import Enum
+from json import JSONDecodeError
+from xml.etree import ElementTree as ET
+from xml.etree.ElementTree import ElementTree, fromstring, tostring
+
+from symphony.bdk.core.service.exception import MessageParserError
+from symphony.bdk.gen.agent_model.v4_message import V4Message
+
+
+def get_text_content_from_message(message: V4Message):
```
thibauult: I need it in my current PR !! :)

thibauult (approved):

symphony/bdk/core/service/message/message_parser.py
```diff
@@ -0,0 +1,103 @@
...
+    to the users mentioned
+
+    :param message: incoming V4 message to be parsed
+    :return: list of users ids that has been mentioned inside the message
+    """
+    mentions_list = _get_tags(message, _EntityTypeEnum.MENTION)
+    return [int(user_id) for user_id in mentions_list]
+
+
+def get_hashtags(message: V4Message):
```
symphony-elias: Please add type hints for the return type.

symphony-elias (approved):
LGTM

symphony/bdk/core/service/message/messageml_util.py
```diff
@@ -0,0 +1,51 @@
...
+    parsed_text = ""
+    matches = re.finditer(_pattern(), raw_text)
+    for match in matches:
+        parsed_text += raw_text[current_parsed_index:match.start()] + _replacement(match.group(0))
+        current_parsed_index = match.end()
+    return parsed_text + raw_text[current_parsed_index:]
+
+
+def _pattern():
+    return "|".join(_special_chars_dic)
```
symphony-youness: This is me being very picky: maybe a t for dicT at the end. _special_chars_dict instead of _special_chars_dic

symphony-mariacristina: I thought it was already called like that! XD

symphony/bdk/core/service/message/message_parser.py
symphony-mariacristina: done in the next commit

symphony/bdk/core/auth/exception.py
```diff
@@ -7,6 +7,7 @@ class AuthInitializationError(Exception):
     """
 
     def __init__(self, message: str):
+        Exception.__init__(self, message)
```
symphony-youness: Curious question: Why do you call the parent constructor for Exception ?
.
.
On internet forums, the usual answer would be `super().__init__` instead of `Exception.__init__` as it would avoid us using the base class name explicitly.

I however don't see the point of calling the parent constructor if we set the instance attributes. 
If you check the source code for Exception:
```Python
class Exception(BaseException):
    """ Common base class for all non-exit exceptions. """
    def __init__(self, *args, **kwargs): # real signature unknown
        pass
```
It only instanciates them for us. So my take would be droping the `super().__init__` completly: https://docs.python.org/3.8/tutorial/errors.html -> 8.5. User-defined Exceptions

symphony/bdk/core/service/message/message_parser.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+def _get_tags(message, entity_type):
+    json_data = _parse_json_data(message.data)
+    tags_list = []
+    for item in json_data.values():
+        try:
+            if item["type"] == entity_type.value:
+                tags_list.append(item["id"][0]["value"])
+        except KeyError:
```
symphony-youri, line 86: if this error case can happen a lot maybe we would be better checking for it explictly (I don't know if it is similar to Java but exceptions have a performance cost and sometimes you want to avoid them)

symphony/bdk/core/service/message/message_parser.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+def get_text_content_from_message(message: V4Message) -> str:
+    """Get text content from PresentationML on incoming messages
+
+    :param message: message containing the PresentationML to be parsed
+    :return: the message text content extracted from the given PresentationML
+    """
+
+    unparsable_presentationml = "Unable to parse the PresentationML, it is not in the correct format."
```
symphony-youri: should we move this to the except block?

symphony-youri (approved):

symphony/bdk/core/service/message/message_parser.py
symphony-mariacristina: The data object coming as payload from the message should be well formed in general. I put this block just to be safe that we are handling this scenario.

symphony/bdk/core/auth/exception.py
symphony-mariacristina: I updated these exceptions following the discussion on this page: https://stackoverflow.com/questions/2901000/the-correct-way-to-define-an-exception-in-python-without-pylint-complaining/2901085#290108 . 
Pylint was complaining and raising warnings about it, so I followed this comment https://stackoverflow.com/a/2901085 and seems like Pylint is happy now. 

But if there's a better way on how to handle this I'm open to change it.

symphony/bdk/core/service/message/message_parser.py
symphony-mariacristina: yes, it makes sense since It is used only there. I put it directly inside the raise exception line

symphony/bdk/core/auth/exception.py
symphony-youness: These discussions seem a bit old (dates of 2010), since python 3+, you would see things more like this: `super().__init__`: https://stackoverflow.com/questions/1319615/proper-way-to-declare-custom-exceptions-in-modern-python

I explained how I see things and how it was documented in the official documentation. Calling the parent constructor is not doing anything anyways which I think we should remove it.
https://docs.python.org/3.8/tutorial/errors.html -> 8.5. User-defined Exceptions

PR #151, opened 2021-03-17
#143 Activity API

### LINKED ISSUE #143: Activity API
Goal is to implement a subclass of [`real_time_event_listener.py`](https://github.com/SymphonyPlatformSolutions/symphony-api-client-python/blob/2.0/symphony/bdk/core/service/datafeed/real_time_event_listener.py#L22) which overrides only on_message_sent method.

Implement:

- [ ] a class AbstractActivity with a matcher and on_activity methods.
- [ ] a CommandClass which reacts to on_message_sent only and parses the text inside the received PresentationML
- [ ] a PatternCommand class (subclass of the above) which filters based on a regex
- [ ] a SlashCommand class (subclass of the above) which filters based on a command and optional bot mention.

Take inspiration from the existing [Java BDK Activity API](https://github.com/finos/symphony-bdk-java/blob/main/docs/activity-api.md).

thibauult:
### Ticket
Closes #143

### Description
Base classes for the Activity API. I think that we won't be able to backport everything from the [Java one](https://github.com/finos/symphony-bdk-java/blob/main/docs/activity-api.md), such as the [Slash Command](https://github.com/finos/symphony-bdk-java/blob/main/docs/activity-api.md#slash-command). Maybe we actually don't have to. 

Is also implemented: the `SessionService` that was required in order to retrieve the bot display name (bot is mention or not to trigger a command). 

Unit tests and docs are coming...

### Checklist

### DIFF
--- a/docsrc/markdown/activity-api.md
+++ b/docsrc/markdown/activity-api.md
@@ -1,0 +1,77 @@
+# Activity API
+
+The Activity API is an abstraction built on top of the Datafeed's [_Real Time Events_](https://developers.symphony.com/restapi/docs/real-time-events). An Activity is basically a user interaction triggered from the chat.
+Two different kinds of activities are supported by the BDK:
+- **Command Activity**: triggered when a message is sent in an `IM`, `MIM` or `Chatroom`
+- **Form Activity**: triggered when a user replies to an [_Elements_](https://developers.symphony.com/symphony-developer/docs/overview-of-symphony-elements) form message
+
+## Activity Registry
+The central component for activities is the [`ActivityRegistry`](../../symphony/bdk/core/activity/registry.py).
+This component is used to either add or retrieve activities. It is accessible from the `SymphonyBdk` object.
+
+```python
+from symphony.bdk.core.symphony_bdk import SymphonyBdk
+from symphony.bdk.core.config.loader import BdkConfigLoader
+
+async def run():
+  async with SymphonyBdk(BdkConfigLoader.load_from_symphony_dir("config.yaml")) as bdk:
+    activity_registry = bdk.activities()
+```
+
+## Command Activity
+A command activity is triggered when a message is sent in an `IM`, `MIM` or `Chatroom`. This is the most basic interaction 
+between an end-user and the bot. Here are some command activity examples: 
+
+- the bot is mentioned followed by a _slash_ command:
+```
+$ @BotMention /buy
+```
+- a command with parameters, the bot is not mentioned:
+```
+$ /buy 1000 goog
+```
+- any message that contains 'hello' can be a command:
+```
+$ I want to say hello to the world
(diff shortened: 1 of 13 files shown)


symphony/bdk/core/activity/api.py
```diff
@@ -0,0 +1,55 @@
...
+        return self._source_event
+
+
+class AbstractActivity(ABC, Generic[C]):
+    """
+    Base abstract class for activities provided by the BDK. Provides a generic flow to process an incoming chat event.
+    """
+
+    @abstractmethod
+    def matches(self, context: C) -> bool:
```
symphony-youness, line 47: Maybe we should add docstrings for the methods of the `AbstractActivity` class

symphony-youness (approved):
LGTM

symphony/bdk/core/activity/api.py
thibauult: Now done via 669a8c1

symphony-elias (approved):
LGTM !

docsrc/markdown/activity-api.md
```diff
@@ -0,0 +1,77 @@
...
+  async with SymphonyBdk(BdkConfigLoader.load_from_symphony_dir("config.yaml")) as bdk:
+    await bdk.activities().register(HelloCommandActivity(bdk.messages()))
+    await bdk.datafeed().start()
+
+
+class HelloCommandActivity(CommandActivity):
+
+  def __init__(self, messages: MessageService):
+    self._messages = messages
+    super().__init__()
```
symphony-youri, line 61: in Python is it ok to not call init first?

docsrc/markdown/activity-api.md
```diff
@@ -0,0 +1,77 @@
...
+  
+  def matches(self, context: CommandContext) -> bool:
```
symphony-youri, line 63: is there something we could with decorator? 
like having a @command("/hello") somewhere?

symphony-youri (commented):
It feels a bit we are translating the Java BDK in Python sometime unfortunately that's not easy for us to be more pythonesque given our expertise :)
maybe before the 2.0 release we should take a bit of time, read code from the most popular Python frameworks and see if there are some principles we could apply?
(for instance I was looking at Flask and they use decorators for controller's routes)

docsrc/markdown/activity-api.md
thibauult: I created https://perzoinc.atlassian.net/browse/PLAT-10706

docsrc/markdown/activity-api.md
thibauult: That's what I understood from https://stackoverflow.com/questions/2399307/how-to-invoke-the-super-constructor-in-python

PR #152, opened 2021-03-18
PLAT-10646: Unknown event breaks DF loop

symphony-youri:
### Ticket
[PLAT-10646](https://perzoinc.atlassian.net/browse/PLAT-10646)

### Description
An unknown event was causing the handle_events part of the DF loop to
exit (i.e return instead of continuing the loop).

Also print the event type in the log to have it with context.

### Dependencies
N/A

### Checklist
- [-] Docstrings added or updated
- [-] Updated the documentation in [docs folder](../docs)


symphony-elias (approved):
LGTM

PR #154, opened 2021-03-19
Backport changes from main to 1.3-rc

symphony-youri:
Preparing the 1.3.4 release

Mostly to include #152 but also taking latest small documentation changes from master.


symphony-elias:
We should change the version here: https://github.com/SymphonyPlatformSolutions/symphony-api-client-python/blob/master/setup.py#L12

symphony-youri:
Of course we should! 🙈

symphony-youness (approved):

PR #158, opened 2021-03-23
PLAT-10588: Implemented concurrent DF loop

symphony-elias:
### Ticket
PLAT-10588

### Description
Implemented concurrency of event handlings in the datafeed loop.
Refactored the DF loops along with unit tests.

### Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,7 +17,7 @@ urllib3 = "^1.26.2"
 aiohttp = "^3.7.3"
 pyyaml = "^5.3.1"
 python-jose = "^3.2.0"
-cryptography = "ˆ3.4.6"
+cryptography = "^3.4.6"
 
 [tool.poetry.dev-dependencies]
 pytest = "^6.2"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,7 @@ python-dateutil = "^2.8.1"
 urllib3 = "^1.26.2"
 aiohttp = "^3.7.3"
 pyyaml = "^5.3.1"
-python-jose = "^3.2.0"
+PyJWT= "^2.0.1"
 cryptography = "^3.4.6"
 
 [tool.poetry.dev-dependencies]
--- a/symphony/bdk/core/service/datafeed/abstract_datafeed_loop.py
+++ b/symphony/bdk/core/service/datafeed/abstract_datafeed_loop.py
@@ -73,15 +73,13 @@ class AbstractDatafeedLoop(ABC):
         self.bdk_config = config
         self.api_client = datafeed_api.api_client
         self.running = False
+        self.hard_kill = False
+        self.timeout = None
         self.tasks = []
 
-    async def start(self, hard_kill: bool = False, timeout: float = None):
+    async def start(self):
         """Start the datafeed event service
 
-        :param hard_kill: if set to True, tasks running listener methods will be cancelled immediately. Otherwise, tasks
-          will be awaited until completion.
-        :param timeout: timeout in seconds to wait for tasks completion when loop stops.
-          None means wait until completion. Ignored if hard_kill set to True.
         :return: None
         """
         logger.debug("Starting datafeed loop")
@@ -91,13 +89,19 @@ class AbstractDatafeedLoop(ABC):
             await self._run_loop()
         finally:
             logger.debug("Stopping datafeed loop")
-            await self._stop_listener_tasks(hard_kill, timeout)
+            await self._stop_listener_tasks()
 
-    async def stop(self):
+    async def stop(self, hard_kill: bool = False, timeout: float = None):
         """Stop the datafeed event service
 
+        :param hard_kill: if set to True, tasks running listener methods will be cancelled immediately. Otherwise, tasks
+          will be awaited until completion.
+        :param timeout: timeout in seconds to wait for tasks completion when loop stops.
+          None means wait until completion. Ignored if hard_kill set to True.
         :return: None
         """
+        self.hard_kill = hard_kill
+        self.timeout = timeout
         self.running = False
 
     @abstractmethod
@@ -151,11 +155,11 @@ class AbstractDatafeedLoop(ABC):
             else:
                 raise exc
 
-    async def _stop_listener_tasks(self, hard_kill, timeout):
-        if hard_kill:
+    async def _stop_listener_tasks(self):
+        if self.hard_kill:
             await self._cancel_tasks()
         else:
-            await self._wait_for_completion_or_timeout(timeout)
+            await self._wait_for_completion_or_timeout()
 
     async def _cancel_tasks(self):
         logger.debug("Cancelling %s listener tasks", len(self.tasks))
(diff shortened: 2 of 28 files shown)


pyproject.toml
```diff
@@ -17,7 +17,7 @@ urllib3 = "^1.26.2"
 aiohttp = "^3.7.3"
 pyyaml = "^5.3.1"
 python-jose = "^3.2.0"
-cryptography = "ˆ3.4.6"
+cryptography = "^3.4.6"
```
symphony-elias: Wrong copy-paste 😖

symphony/bdk/core/service/datafeed/abstract_datafeed_loop.py
```diff
@@ -97,12 +139,18 @@ async def handle_v4_event_list(self, events: List[V4Event]):
...
         :param events: the list of the received datafeed events.
         """
+        if not events:
+            return
+
         for event in filter(lambda e: e is not None, events):
             for listener in self.listeners:
                 if await listener.is_accepting_event(event, self.bdk_config.bot.username):
-                    await self._dispatch_on_event_type(listener, event)
+                    asyncio.create_task(self._dispatch_on_event_type(listener, event))
```
symphony-youri: do we have to do something in term of error handling? I remember that is the legacy BDK there were done callbacks used
if an exception is raised in the task it won't break the DF loop or anything?

symphony-youri: to support the soft/hard kill will we have to collect the tasks somewhere so we can cancel() all of the running ones/

tests/core/service/datafeed/abstract_datafeed_loop_test.py
```diff
@@ -0,0 +1,480 @@
...
+            if self.count == 1:
+                await self.queue.get()
+                await bare_df_loop.stop()
+            elif self.count == 2:
+                await self.queue.put("message")
+
+    bare_df_loop.read_datafeed.side_effect = read_df_side_effect
+    bare_df_loop.subscribe(QueueListener())
+
+    await bare_df_loop.start()  # test no deadlock
```
symphony-youri, line 480: is there a way to set a maximum timeout for a python test so that in case of deadlock we don't block the CI build?

symphony-youri: not out of the box maybe (https://pypi.org/project/pytest-timeout/)
not sure it is worth adding an extra library

symphony-youri: maybe https://stackoverflow.com/questions/59073731/set-default-timeout-on-github-action-pipeline is simpler

symphony/bdk/core/service/datafeed/abstract_datafeed_loop.py
symphony-elias: If an exception is raised inside the listener, then asyncio logs it for us:
```
2021-03-24 10:01:58,850 - root - DEBUG - iIaUeR - Received event: Ojwe2s5mMfJbHPlLbEZRUH___oech-ZubQ
2021-03-24 10:02:03,208 - asyncio - ERROR -  - Task exception was never retrieved
future: <Task finished name='Task-13' coro=<AbstractDatafeedLoop._dispatch_on_event_type() done, defined at /Users/elias.croze/workspace/IdeaProjects/symphony-api-client-python/symphony/bdk/core/service/datafeed/abstract_datafeed_loop.py:150> exception=ValueError('error inside listener')>
Traceback (most recent call last):
  File "/Users/elias.croze/workspace/IdeaProjects/symphony-api-client-python/symphony/bdk/core/service/datafeed/abstract_datafeed_loop.py", line 163, in _dispatch_on_event_type
    await listener_method(event.initiator, event_field)
  File "/Users/elias.croze/workspace/IdeaProjects/symphony-api-client-python/examples/datafeed.py", line 28, in on_message_sent
    raise ValueError("error inside listener")
ValueError: error inside listener
```
IMHO this is enough.

symphony-elias: To retrieve unfinished tasks, we can use: https://docs.python.org/3/library/asyncio-task.html#asyncio.all_tasks

symphony-youri: is there a risk we will retrieve user's define tasks and stop them (like if the bot is doing something else beside running the DF loop)?

symphony-elias: I added a list with the listener tasks. Please bear in mind lists are thread safe: https://stackoverflow.com/a/6319267

symphony/bdk/core/service/datafeed/abstract_datafeed_loop.py
```diff
@@ -72,30 +73,30 @@ def __init__(self, datafeed_api: DatafeedApi, auth_session: AuthSession, config:
...
+
+    async def stop(self, hard_kill=False):
         """Stop the datafeed event service
 
+        :param hard_kill: if set to True, tasks running listener methods will be cancelled immediately. Otherwise, tasks
+          will be awaited until completion.
         :return: None
         """
         self.running = False
+        self.hard_kill = hard_kill
```
symphony-youri, line 103: is there a chance that the loop stops before we set the hard_kill value?

symphony/bdk/core/service/datafeed/abstract_datafeed_loop.py
```diff
@@ -72,30 +73,30 @@ def __init__(self, datafeed_api: DatafeedApi, auth_session: AuthSession, config:
...
-    async def stop(self):
+        await self._run_loop()
+
+        logger.debug("Stopping datafeed loop")
+        await self._stop_listener_tasks()
+
+    async def stop(self, hard_kill=False):
         """Stop the datafeed event service
 
+        :param hard_kill: if set to True, tasks running listener methods will be cancelled immediately. Otherwise, tasks
```
symphony-youri, line 97: if we have a pending request to get events we still have to wait for it to timeout? (if so maybe we can document this as well)

symphony/bdk/core/service/datafeed/abstract_datafeed_loop.py
symphony-elias: Loop can stop if there is an exception inside the `while` loop, typically when calling `await self.read_datafeed()` if it throws a non 400 error. In this case, exception is propagated.
Indeed, we have to handle that case.

symphony/bdk/core/service/datafeed/abstract_datafeed_loop.py
symphony-elias: If `hard_kill` is set to `False`, we wait forever until tasks are completed.

symphony/bdk/core/service/datafeed/abstract_datafeed_loop.py
symphony-elias: Refactored the loop: `self.hard_kill` and `self.timeout` updated before `self.running`, finally block added

symphony/bdk/core/service/datafeed/abstract_datafeed_loop.py
symphony-elias: Timeout option added

symphony-youri (approved):

PR #161, opened 2021-03-31
PLAT-10563: FormReply activity

symphony-youness:
### Ticket
[PLAT-10563](https://perzoinc.atlassian.net/browse/PLAT-10563)

### Description
Base classes for a FormReply Activity.

Added a quick fix in the ActivityRegistry: The problem was that the `act.before_matcher()` method was called before checking the correct instance of the activity. 

Documentation is coming ...

### Checklist

### DIFF
--- a/docsrc/markdown/activity-api.md
+++ b/docsrc/markdown/activity-api.md
@@ -77,6 +77,7 @@ except KeyboardInterrupt:
 ```
 1. The `matches()` method allows the activity logic to be triggered when a message starts by a mention to the bot and the text `/hello` separated by a space. Ex: `@bot_name /hello`
 2. The activity logic. Here, we send a message: "Hello, World"
+
 ## Form Activity
 A form activity is triggered when an end-user replies or submits an Elements form.
 
@@ -99,7 +100,7 @@ For this example, we will assume that the following Elements form has been poste
 2. the form has one unique `<text-field>` called "**name**"
 3. the has one action button called "**submit**"
 
-The following example code handle the above form submission:
+The following code example handles the above form submission:
 ```Python
 import asyncio
 import logging.config
@@ -142,3 +143,22 @@ except KeyboardInterrupt:
 1. The `matches()` method allows the activity logic to be triggered when a form with the `id` `"hello-form"` has been submitted from the action button `"submit"`
 2. The activity context allows us to retrieve form values. Here, the content of the `<text-field>` called "**name**"
 3. We send back the message: "Hello, **text_field_content**"
+
+---
+**NOTE - General activities use case**
+
+For you to not overload your matches method, you can pre-process a context before the `matches()` method is called by
+overriding the `before_matcher()` method of any activity.
+```Python
+class DummyActivityExample(CommandActivity):
+    
+    def matches(self, context: CommandContext) -> bool:
+        return context.some_attribute == "some_value"
+    
+    async def on_activity(self, context: CommandContext):
+        # Do stuff
+
+    def before_matcher(self, context: CommandContext):
+        context.some_attribute = some_call_to_another_service()
+```
+---
--- a/docsrc/markdown/activity-api.md
+++ b/docsrc/markdown/activity-api.md
@@ -61,10 +61,10 @@ class HelloCommandActivity(CommandActivity):
     super().__init__()
   
   def matches(self, context: CommandContext) -> bool:
-     return context.text_content.startswith("@" + context.bot_display_name + " /hello")
+     return context.text_content.startswith("@" + context.bot_display_name + " /hello") # (1)
   
   async def on_activity(self, context: CommandContext):
-      await self._messages.send_message(context.stream_id, "<messageML>Hello, World!</messageML>")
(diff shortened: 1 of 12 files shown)


docsrc/markdown/activity-api.md
```diff
@@ -75,3 +75,70 @@ try:
 except KeyboardInterrupt:
     logging.info("Ending activity example")
 ```
+1. The `matches()` method allows the activity logic to be triggered when a message starts by a mention to the bot and the text `/hello` separated by a space. Ex: `@bot_name /hello`
+2. The activity logic. Here, we send a message: "Hello, World"
+## Form Activity
```
symphony-youness, line 81: Will add a break on top of this.

examples/activities/activity.py
```diff
@@ -27,8 +26,7 @@ async def on_activity(self, context: CommandContext):
         await self._messages.send_message(context.stream_id, "<messageML>Hello, World!</messageML>")
 
 
-logging.config.fileConfig(os.path.dirname(os.path.abspath(__file__)) + '/logging.conf',
-                          disable_existing_loggers=False)
+logging.config.fileConfig(Path("../logging.conf"), disable_existing_loggers=False)
```
symphony-elias: Beware, I am not sure this will work if you run the script from somewhere else than the script folder.

symphony/bdk/core/activity/form.py
```diff
@@ -0,0 +1,50 @@
...
+    """
+
+    def matches(self, context: FormReplyContext) -> bool:
+        pass
+
+    def on_activity(self, context: FormReplyContext):
+        pass
+
+    def before_matcher(self, context: FormReplyContext):
+        context._form_id = context.source_event.form_id
```
symphony-elias: I'd prefer `context._form_id` and `context._form_values` to be set in the constructor of `FormReplyContext` directly

examples/activities/form_activity.py
```diff
@@ -0,0 +1,56 @@
...
+    async with SymphonyBdk(BdkConfigLoader.load_from_symphony_dir("config.yaml")) as bdk:
+        await bdk.activities().register(SlashGifCommandActivity(bdk.messages()))
+        await bdk.activities().register(ReplyFormReplyActivity(bdk.messages()))
+        await bdk.datafeed().start()
+
+
+class SlashGifCommandActivity(CommandActivity):
+    def __init__(self, messages: MessageService):
+        self._messages = messages
+        super().__init__()
```
symphony-youri: why do we call init here and not in the ReplyFormReplyActivity?

symphony/bdk/core/activity/form.py
symphony-youri: as the event is coming from the outside should we be defensive about what we get? (like if it is null or does not have a form id)

symphony-youri: > I'd prefer `context._form_id` and `context._form_values` to be set in the constructor of `FormReplyContext` directly

true, I wonder why we have the before_matcher() if we can do part of the context's constructor

tests/core/activity/registry_test.py
```diff
@@ -62,3 +73,16 @@ async def test_on_message_sent(activity_registry, message_sent, command):
 
     command.before_matcher.assert_called_once()
     command.matches.assert_called_once()
+
+
+@pytest.mark.asyncio
+async def test_on_symphony_elements_action(activity_registry, elements_action, form):
```
symphony-youri, line 116: maybe we could add tests to cover different cases like the event not being an ElementsAction
and the matcher not returning true
(right now we have 100% lines coverage but not 100% branch coverage)

examples/activities/form_activity.py
symphony-youness: **I will remove this along with the constructor of the `CommandActivity`**

```Python
def __init__(self):
    self.bot_display_name = str
```
This was done in java and I think it's purpose was to set the bot name in the activity.
ActivityRegistery.java : `a command activity (potentially) needs the bot display name in order to parse the message text content` (maybe we don't actually need it)

tests/core/activity/registry_test.py
symphony-youness: will do

symphony/bdk/core/activity/form.py
symphony-youness: I tried sending a form without an `id` and It didn't work: ` {"code":400,"message":"Failed to parse MessageML: The attribute \"id\" is required"} so is that possible ?

I thought it was to separate the logic of modifying the context in `before_matcher()` before validating it in `matches()`

To handle the possible unwanted values we get, maybe we can do this, WDT @symphony-elias :
```Python
self._form_id = getattr(source_event, "form_id", None)
or
self._form_id = getattr(source_event, "form_id", "")
```
We definitely can add those to the constructor. Maybe I am missing the purpose of thibault setting them in `before_matcher()`

examples/activities/activity.py
symphony-youness: Indeed, good catch. Will fix

symphony-youri (approved):

PR #162, opened 2021-03-31
PLAT-10652: regenerated all models to enable autocomplete

symphony-elias:
### Ticket
PLAT-10562

### Description
Regenerated all models to have fields declared with type hints in classes
First commit https://github.com/SymphonyPlatformSolutions/symphony-api-client-python/pull/162/commits/f8639831f9b6cbbd514ceefb7b3050db1f8f2c69 contains the updates on generated files
Second commit https://github.com/SymphonyPlatformSolutions/symphony-api-client-python/pull/162/commits/7d09bc58ae593ea148207e7af023086ddfa5ae40 contain the script file to generate the model and api files and the documentation on how to do that.
Last commit https://github.com/SymphonyPlatformSolutions/symphony-api-client-python/pull/162/commits/836ef156ec17bf4fcc1ac1047ce7f84a6633bc1b is a change in the ApiClient to ignore unknown fields.

### Dependencies
⚠️ Generated with a local version of the generator: https://github.com/symphony-elias/openapi-generator/compare/master...symphony-elias:PLAT-10652?expand=1

### Checklist

### DIFF
--- a/symphony/bdk/core/client/api_client_factory.py
+++ b/symphony/bdk/core/client/api_client_factory.py
@@ -69,7 +69,7 @@ class ApiClientFactory:
         await self._session_auth_client.close()
 
     def _get_api_client(self, server_config, context) -> ApiClient:
-        configuration = Configuration(host=(server_config.get_base_path() + context))
+        configuration = Configuration(host=(server_config.get_base_path() + context), discard_unknown_keys=True)
         configuration.verify_ssl = True
         configuration.ssl_ca_cert = self._config.ssl.trust_store_path
 
--- a/docsrc/markdown/tech/generated_api.md
+++ b/docsrc/markdown/tech/generated_api.md
@@ -9,8 +9,11 @@ from [swagger specifications](https://github.com/symphonyoss/symphony-api-spec) 
 * [login/login-api-public.yaml](https://github.com/symphonyoss/symphony-api-spec/blob/master/login/login-api-public.yaml)
 * [pod/pod-api-public-deprecated.yaml](https://github.com/symphonyoss/symphony-api-spec/blob/master/pod/pod-api-public-deprecated.yaml)
 
-In order to re-generate these files:
-* download the [openapi-generator-cli jar](https://search.maven.org/artifact/org.openapitools/openapi-generator-cli/5.0.0/jar)
-  from Maven Central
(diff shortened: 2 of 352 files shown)


symphony-youri:
I'm a little worried about maintaining a copy of the openapi generator

Is 
```
public String getDataTypeHint() {
        if(this.dataType.startsWith("[") && this.dataType.endsWith("]")) {
            return "List" + this.dataType;
        }
        return this.dataType;
    }
```

something we could with Mustache?

But then about the imports I don't know what we could do

symphony-elias:
> I'm a little worried about maintaining a copy of the openapi generator
> 
> Is
> 
> ```
> public String getDataTypeHint() {
>         if(this.dataType.startsWith("[") && this.dataType.endsWith("]")) {
>             return "List" + this.dataType;
>         }
>         return this.dataType;
>     }
> ```
> 
> something we could with Mustache?
> 
> But then about the imports I don't know what we could do

No, we cannot do that with Mustache (Mustache is logic-less so no "if") and if we don't put `List[Type]`, autocompletion does not work.
There is an Handlebars engine in the openapi-generator that can be activated with `-e handlebars` option but I didn't manage to make it work because of jmustache specific keywords (`-first` and `-last`).

About the imports, the circular import issue arises because of the use of discriminator in the swagger specs. And because we don't want to use lazy imports (which makes autocomplete not working)...

symphony-youri (approved):

symphony/bdk/core/client/api_client_factory.py
```diff
@@ -69,7 +69,7 @@ async def close_clients(self):
         await self._session_auth_client.close()
 
     def _get_api_client(self, server_config, context) -> ApiClient:
-        configuration = Configuration(host=(server_config.get_base_path() + context))
+        configuration = Configuration(host=(server_config.get_base_path() + context), discard_unknown_keys=True)
```
symphony-youness, line 72: I had already created a no pointer task to add this: [PLAT-10751](https://perzoinc.atlassian.net/browse/PLAT-10751)
Will close it as fixed by this one.

symphony-youness (approved):

symphony/bdk/core/client/api_client_factory.py
symphony-elias: Yes, because datafeed example was failing because of that (new stateful flag added for VSM).

PR #168, opened 2021-04-07
PLAT-10555: Deserializing models in tests

symphony-youness:
### Ticket
[PLAT-10555](https://perzoinc.atlassian.net/browse/PLAT-10555)

### Description
We update the tests to use the generated models instead of creating objects from a json file.
This way we keep alignment with model attribute names in tests and the exposed models.

There are some problems that were identified and I will point them out in the conversation.

### Checklist

### DIFF
--- a/tests/core/service/application/application_service_test.py
+++ b/tests/core/service/application/application_service_test.py
@@ -11,7 +11,7 @@ from symphony.bdk.gen.pod_model.pod_app_entitlement import PodAppEntitlement
 from symphony.bdk.gen.pod_model.pod_app_entitlement_list import PodAppEntitlementList
 from symphony.bdk.gen.pod_model.user_app_entitlement import UserAppEntitlement
 from symphony.bdk.gen.pod_model.user_app_entitlement_list import UserAppEntitlementList
-from tests.utils.resource_utils import object_from_json_relative_path, object_from_json
+from tests.utils.resource_utils import object_from_json, get_deserialized_object_from_resource
 
 
 @pytest.fixture(name="auth_session")
@@ -42,7 +42,7 @@ def fixture_application_service(application_api, app_entitlement_api, auth_sessi
 async def test_create_application(application_api, application_service):
     application_api.v1_admin_app_create_post = AsyncMock()
     application_api.v1_admin_app_create_post.return_value = \
-        object_from_json_relative_path("application/create_application.json")
+        get_deserialized_object_from_resource(ApplicationDetail, "application/create_application.json")
 
     application_detail = await application_service.create_application(ApplicationDetail())
 
@@ -51,7 +51,7 @@ async def test_create_application(application_api, application_service):
         application_detail=ApplicationDetail()
     )
 
-    assert application_detail.applicationInfo.appId == "my-test-app"
+    assert application_detail.application_info.app_id == "my-test-app"
     assert application_detail.description == "a test app"
 
 
@@ -59,7 +59,7 @@ async def test_create_application(application_api, application_service):
 async def test_update_application(application_api, application_service):
     application_api.v1_admin_app_id_update_post = AsyncMock()
     application_api.v1_admin_app_id_update_post.return_value = \
-        object_from_json_relative_path("application/update_application.json")
+        get_deserialized_object_from_resource(ApplicationDetail, "application/update_application.json")
 
     application_detail = await application_service.update_application("my-test-app", ApplicationDetail())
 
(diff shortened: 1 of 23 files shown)


tests/core/service/application/application_service_test.py
```diff
@@ -145,14 +145,16 @@ async def test_update_application_entitlements(app_entitlement_api, application_
     )
 
     assert len(pod_app_entitlements) == 1
-    assert pod_app_entitlements[0].appId == "rsa-app-auth-example"
+    assert pod_app_entitlements[0].app_id == "rsa-app-auth-example"
 
 
+@pytest.mark.skip(
```
symphony-youness, line 151: This test is skipped because the payload used to test it raises an error with the generated model.

UserAppEntitlement -> ProductList -> Product -> `sku` is required 

The payload is the same in the [documentation](https://developers.symphony.com/restapi/reference#user-apps) and the [java BDK](https://github.com/finos/symphony-bdk-java/blob/main/symphony-bdk-core/src/test/resources/application/user_apps.json) and one of the products doesn't have the `sku` attribute.

In python, the [Product](https://github.com/SymphonyPlatformSolutions/symphony-api-client-python/blob/2.0/symphony/bdk/gen/pod_model/product.py) model takes `sku` as a required argument. [Api spec -> product](https://github.com/symphonyoss/symphony-api-spec/blob/master/pod/pod-api-public.yaml#L5353) describes `sku` as required 

Solution: Add the "sku" in the test payload. Does that mean the documentation is wrong? Do we care? How to update it?

tests/utils/resource_utils.py
```diff
@@ -27,11 +28,22 @@ def object_from_json_relative_path(relative_path):
...
+    """Deserializes the passed payload to an instance of the specified model
+    Disregards unknown fields is
+
+    :param model: OpenApi generated model
+    :param payload: json payload to be deserialized
+    :return: Instance of the model
+    """
+    response = namedtuple("MockResp", ["status", "reason"])(200, "")
+    response = RESTResponse(response, payload)
+    return ApiClient(configuration=Configuration(discard_unknown_keys=True)).deserialize(response, (model,), True)
```
symphony-youness: The `configuration=Configuration(discard_unknown_keys=True)` is a tricky one.

Not using it in the core code of the ApiClient() raises an issue with the [list-audit-trail](https://developers.symphony.com/restapi/reference#list-audit-trail-v1) endpoint: The payload for this endpoint returns multiple attributes that are not described in the [API spec](https://github.com/symphonyoss/symphony-api-spec/blob/master/agent/agent-api-public.yaml#L5144) which means if someone uses this endpoint right now, it will raise an Exception.

Same problem for the [create-app](https://developers.symphony.com/restapi/reference#create-app) endpoint.

Conclusion: We need to add it to the ApiClient in the BDK for the list-audit-trail and create-app endpoint to work.

tests/core/service/application/application_service_test.py
symphony-elias: @symphony-youri @symphony-soufiane  Do you have any clue ? IMHO documentation and swagger specs should be consistent with actual behavior

tests/utils/resource_utils.py
symphony-elias: In the java BDK unknown properties are ignored: https://github.com/finos/symphony-bdk-java/blob/70745c703147d1ba08eeaacad18868c18c237091/symphony-bdk-http/symphony-bdk-http-jersey2/src/main/java/com/symphony/bdk/http/jersey2/JSON.java#L23
So we could set `Configuration(discard_unknown_keys=True)` in the `ApiClientFactory` for all clients and open a bug ticket for the endpoints whose behavior is inconsistent with documentation and swagger specs.

tests/core/service/application/application_service_test.py
symphony-youri: I'd say we should update the spec to make it non required, this needs to be done in SBE and in swagger spec repos

tests/utils/resource_utils.py
symphony-youri: We definitely want the BDK to be able to support unknown keys, for instance if you are targeting a new version of SBE with new attributes being returned you still want the BDK to work

symphony-youri: do we need a full APIClient to do the deserialization?

symphony-youness: For sure, we don't. Will correct it.

symphony-youness: I will create a ticket to add that configuration.

symphony-youness: I used the underlying method used in ApiClient(). deserialize()

symphony-elias (approved):
Looks good !
Please reference the ticket you created for further bug fixes / improvements/

PR #175, opened 2021-04-15
PLAT-10706: @slash decorator

symphony-elias:
### Ticket
PLAT-10706

### Description
Created a slash decorator to register new slash commands.

Made `bdk.activities().register(command)` synchronous because I could not have the desired effect with an async `register` method. 

### Checklist

### DIFF
--- a/symphony/bdk/core/activity/command.py
+++ b/symphony/bdk/core/activity/command.py
@@ -1,4 +1,5 @@
 import logging
+import re
 
 from symphony.bdk.core.activity.api import AbstractActivity, ActivityContext
 from symphony.bdk.core.activity.exception import FatalActivityExecutionException
@@ -11,8 +12,7 @@ logger = logging.getLogger(__name__)
 
 
 class CommandContext(ActivityContext[V4MessageSent]):
-    """
-    Default implementation of the :py:class:`ActivityContext` handled by the :py:class:`CommandActivity`.
+    """Default implementation of the :py:class:`ActivityContext` handled by the :py:class:`CommandActivity`.
     """
 
     def __init__(self, initiator: V4Initiator, source_event: V4MessageSent, bot_display_name: str):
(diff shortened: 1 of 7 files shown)


examples/datafeed.py
```diff
@@ -3,27 +3,20 @@
...
 from symphony.bdk.core.symphony_bdk import SymphonyBdk
-from symphony.bdk.gen.agent_model.v4_initiator import V4Initiator
-from symphony.bdk.gen.agent_model.v4_message_sent import V4MessageSent
 
 
 async def run():
     config = BdkConfigLoader.load_from_symphony_dir("config.yaml")
 
     async with SymphonyBdk(config) as bdk:
-        datafeed_loop = bdk.datafeed()
```
symphony-elias: I will remove that

symphony/bdk/core/activity/command.py
```diff
@@ -43,12 +43,41 @@ def bot_display_name(self) -> str:
...
+        :param name: the command name
+        :param requires_mention_bot: if the command requires the bot mention to trigger the slash command
+        :param callback: the coroutine to be executed if message text matches
+        """
+        self._name = name
+        self._requires_mention_bot = requires_mention_bot
+        self._callback = callback
+
+    def matches(self, context: CommandContext) -> bool:
+        pattern = rf"@{context.bot_display_name} {self._name}" if self._requires_mention_bot else rf"{self._name}"
```
symphony-elias, line 79: I was wondering: do we have backslashes in bot names ? If so, I will keep raw strings.

examples/datafeed.py
```diff
@@ -3,27 +3,20 @@
...
-        datafeed_loop.subscribe(RealTimeEventListenerImpl())
-        await datafeed_loop.start()
+        @bdk.slash("/hello")
```
symphony-youri: can we use it out of the with block?

symphony-youri: probably not it is using the bdk instance right?

same as

```
from flask import Flask
app = Flask(__name__)

@app.route('/')
def hello_world():
    return 'Hello, World!'
```

symphony/bdk/core/activity/command.py
symphony-youri: The API does not document, I tried to find if there was a check in SBE but I could not find where this is happening
Maybe something to try using ACP?

symphony-youri (commented):
👍

symphony/bdk/core/symphony_bdk.py
```diff
@@ -221,11 +222,28 @@ def presence(self) -> PresenceService:
     def activities(self) -> ActivityRegistry:
         """Get the :class:`ActivityRegistry` from the BDK entry point.
 
-        :return: The :class:`ActivityRegistry instance.
+        :return: The :class:`ActivityRegistry` instance.
 
         """
         return self._activity_registry
 
+    def slash(self, command: str, mention_bot: bool = True):
```
thibauult: Is it possible to move it inside the `activity` module?

examples/datafeed.py
symphony-elias: Yes it is using the bdk instance. But a user could not use a `with` block and manually call the `close_clients` method in a `finally` block.

symphony/bdk/core/symphony_bdk.py
symphony-elias: yes, then the example would go as following:
```python
        activities = bdk.activities()

        @activities.slash("/hello")
        async def callback(context: CommandContext):
            logging.debug("Received slash command: %s", context.text_content)

        await bdk.datafeed().start()

```

symphony/bdk/core/activity/command.py
symphony-elias: Yes I managed to create such a user: https://devx1.symphony.com/admin-console/#user/12987981108841

symphony-youness (approved):
LGTM

symphony-youri (approved):

PR #182, opened 2021-04-30
PLAT-10817: documented how to solve self signed certificate issues

symphony-elias:
### Ticket
PLAT-10817

### Description
Documented how to solve SSL issues

### Checklist
- [N/A] Unit tests updated or added

### DIFF
--- a/docsrc/index.md
+++ b/docsrc/index.md
@@ -12,6 +12,7 @@ Contents
 * [Application service](markdown/application_service.md)
 * [Signal service](markdown/signal_service.md)
 * [Presence service](markdown/presence_service.md)
+* [FAQ](markdown/faq.md)
 
 ```eval_rst
 .. autosummary::
--- a/docsrc/markdown/configuration.md
+++ b/docsrc/markdown/configuration.md
@@ -118,7 +118,7 @@ manager which manages the key token of the bot.
   under the hood, you can check
   [ssl lib documentation on certificates](https://docs.python.org/3/library/ssl.html#certificates) for more information.
 
-  To fetch the cert file in pem format, you can run the following openssl command: `openssl s_client -connect <host>:<port> -showcerts > host.cert`
+  To fetch the cert file in pem format, you can run the following openssl command: `openssl s_client -connect <host>:<port> -showcerts > cert.pem`
 - `bot` contains information about the bot like the username, the private key for authenticating the service account
   on pod.
 - `app` contains information about the extension app that the bot will use like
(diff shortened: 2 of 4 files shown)


docsrc/markdown/known_issues.md
```diff
@@ -0,0 +1,49 @@
+# Known issues
```
symphony-youri: I would not call it an issue but a workaround 😛

docsrc/markdown/known_issues.md
```diff
@@ -0,0 +1,49 @@
...
+```
+
+If you use different hosts and ports for the `agent`, `keyManager` and `sessionAuth`, you will have to repeat the
+command above with the specific hosts and ports and append all the certificates to the same file.
+
+Then, add the following to the configuration:
+
+```yaml
+ssl:
+  trustStore:
```
symphony-youri: Would adding the certificate to the system keychain be a workaround too?

docsrc/markdown/known_issues.md
```diff
@@ -0,0 +1,49 @@
...
+File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/asyncio/sslproto.py", line 189, in feed_ssldata
+self._sslobj.do_handshake()
+
+File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/ssl.py", line 944, in do_handshake
+self._sslobj.do_handshake()
+
+ssl.SSLCertVerificationError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self signed certificate in certificate chain (_ssl.c:1123)
+```
+
+means your pod uses a self-signed certificate.
```
symphony-youri: Does it mean that develop2 is a self signed certificate? It appears ok here: https://www.sslshopper.com/ssl-checker.html#hostname=https://develop2.symphony.com

symphony-elias: This is strange indeed... I will update the sentence.

docsrc/markdown/known_issues.md
symphony-elias: Should FAQ be a better name ?

docsrc/markdown/known_issues.md
symphony-youri: Actually with https://www.ssllabs.com/ssltest/analyze.html?d=develop2.symphony.com&latest the last digicert in the chain is indeed marked as self signed

symphony-youri (approved):
An alternative would be to support an insecure flag (similar to curl -k) but I'm not a big fan of supporting this

PR #184, opened 2021-05-10
PLAT-10695 AsyncIO/Proxy usage on Windows

symphony-mariacristina:
### Ticket
[PLAT-10695](https://perzoinc.atlassian.net/browse/PLAT-10695)

### Description
Default event loop policy has to be changed on Windows + Python 3.8 if a proxy is used. Goal of this PR is to make sure that if these conditions applies our BDK 2.0 examples will still be workingas expected, by setting the event loop policy correctly.

Seems like the _init_.py is executed only once https://stackoverflow.com/questions/39452319/at-which-moment-and-how-often-are-executed-the-init-py-files-by-python

### Checklist

### DIFF
--- a/docsrc/markdown/faq.md
+++ b/docsrc/markdown/faq.md
@@ -52,24 +52,3 @@ ssl:
   trustStore:
     path: /path/to/cert.pem
 ```
-
-## How do I solve aiohttp problem while using proxy on Windows?
-There is an outstanding issue with aiohttp and Python 3.8+ on Windows. This affects our client when you specify a proxy 
-on Windows using Python 3.8+.
-
-Under these specific conditions, the following error might appear while trying to start an async loop:
-```
-Exception: ClientConnectorError
-Cannot connect to host localhost.symphony.com:443 ssl:default [wrong parameter]
-  File "D:\local_d\repodata\testingBot\bot\python\main.py", line 27, in run
-    await datafeed_loop.start()
-  File "D:\local_d\repodata\testingBot\bot\python\main.py", line 49, in <module>
-    asyncio.run(run())
-```
-
-If you are using a proxy, make sure to implement the following workaround before you start the async datafeed or 
-make any async calls with aiohttp:
-```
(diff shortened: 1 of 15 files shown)


symphony-youri (commented):
Is there a way we could have this automatically?

like making it part of the imported module definition? so that it runs before starting the event loop but automatically

symphony-youri:
> Is there a way we could have this automatically?
> 
> like making it part of the imported module definition? so that it runs before starting the event loop but automatically

that's tricky because we have to load the proxy first
is there a risk in using this event loop all the time on Windows? (so we drop the dependency on the configuration)

symphony-mariacristina:
> is there a risk in using this event loop all the time on Windows? (so we drop the dependency on the configuration)

I tested by always using `asyncio.WindowsSelectorEventLoopPolicy()` on windows and without a proxy seems like everything is working fine. I'll remove the `config.proxy is not None ` condition.

symphony-youri:
> > is there a risk in using this event loop all the time on Windows? (so we drop the dependency on the configuration)
> 
> I tested by always using `asyncio.WindowsSelectorEventLoopPolicy()` on windows and without a proxy seems like everything is working fine. I'll remove the `config.proxy is not None ` condition.

I remember that I tried to understand the difference between the 2 policies but could not find something very interesting, IMHO Windows is (should?) mostly used for dev so we should be good 🤞

docsrc/markdown/faq.md
```diff
@@ -51,3 +52,4 @@ ssl:
   trustStore:
     path: /path/to/cert.pem
 ```
+
```
symphony-elias: As mentioned in https://github.com/SymphonyPlatformSolutions/generator-symphony/pull/115, it could be great to have a point mentioning the windows + proxy case, the error and the fix.

symphony-youri (commented):
As discussed right now, let's drop the FAQ part since the workaround is embedded now
and add a comment in __init__.py to explain why we do that

symphony-youri (approved):

symphony-youness (approved):

PR #187, opened 2021-05-12
PLAT-10829: Added a Message class to make the sending of message easier

symphony-elias:
### Ticket
PLAT-10829

### Description
Added a Message class to make the sending of message easier
Documentation is missing, to be added when correct naming and method deprecation (or not) is agreed upon.

### Checklist

### DIFF
--- a/symphony/bdk/core/service/message/message_service.py
+++ b/symphony/bdk/core/service/message/message_service.py
@@ -1,5 +1,4 @@
-from io import IOBase
-from typing import Union, List, Tuple
+from typing import Union, List, Tuple, IO
 
 from symphony.bdk.core.auth.auth_session import AuthSession
 from symphony.bdk.core.config.model.bdk_retry_config import BdkRetryConfig
@@ -42,7 +41,7 @@ class OboMessageService:
             message: Union[str, Message],
             data=None,
             version: str = "",
-            attachment: List[Union[IOBase, Tuple[IOBase, IOBase]]] = None
+            attachment: List[Union[IO, Tuple[IO, IO]]] = None
     ) -> V4Message:
         """Send a message to an existing stream.
         See: `Create Message <https://developers.symphony.com/restapi/reference#create-message-v4>`_
@@ -73,8 +72,8 @@ class OboMessageService:
             message: str,
             data: str = "",
             version: str = "",
-            attachment: List[IOBase] = None,
-            preview: List[IOBase] = None
+            attachment: List[IO] = None,
+            preview: List[IO] = None
     ) -> V4Message:
         """Send a message to an existing stream.
         See: `Create Message <https://developers.symphony.com/restapi/reference#create-message-v4>`_
@@ -183,7 +182,7 @@ class MessageService(OboMessageService):
             message: Union[str, Message],
             data=None,
             version: str = "",
-            attachment: List[Union[IOBase, Tuple[IOBase, IOBase]]] = None
+            attachment: List[Union[IO, Tuple[IO, IO]]] = None
     ) -> V4MessageBlastResponse:
         """Send a message to multiple existing streams.
         See: `Blast Message <https://developers.symphony.com/restapi/reference#blast-message>`_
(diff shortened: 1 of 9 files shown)


symphony-elias:
Asking for comments:

- do we want to deprecate the `send_message` and `blast_message` endpoints (with all the parameters) and make it internal, like `_send_message` and `_blast_message` ?
- since there is no method overloading in python, what could be a better naming than `send_simple_message` and `send_complex_message` ?

symphony/bdk/core/service/message/message_service.py
```diff
@@ -34,6 +35,38 @@ def __init__(self, messages_api: MultiAttachmentsMessagesApi, message_suppressio
         self._auth_session = auth_session
         self._retry_config = retry_config
 
+    async def send_simple_message(
```
symphony-youri: would the pythonic way be to have a simple send_message that supports both message as str | Message?

symphony/bdk/core/service/message/message_service.py
```diff
@@ -34,6 +35,38 @@ def __init__(self, messages_api: MultiAttachmentsMessagesApi, message_suppressio
...
+        :param stream_id: The ID of the stream to send the message to.
+        :param message: the message to be sent
+
+        :return: a V4Message object containing the details of the sent message.
+        """
+        return await self.send_message(stream_id, message.content, message.data, message.version, message.attachments,
+                                       message.previews)
+
     @retry
     async def send_message(
```
symphony-youri: don't we want data to be a dict/object instead of a string?
and could we make attachment/preview a list of tuples? [(att1, prev1),...] ?

symphony/bdk/core/service/message/message_service.py
symphony-elias: It seems pythonic way seems to have all arguments put as optional and implement the behavior based on the passed arguments, see https://stackoverflow.com/a/6434546

symphony-elias: One of the question I had is whether we allow ourselves to break the backward compatibility of our API. I would say no, so as suggested, message will be of type str or Message and behavior will be discriminated based on the actual type.

symphony-youri: as we have not published a 2.0 yet we can right?

symphony/bdk/core/service/message/message_service.py
```diff
@@ -135,25 +167,54 @@ async def list_messages(
...
+            "session_token": await self._auth_session.session_token,
+            "key_manager_token": await self._auth_session.key_manager_token,
+            "skip": skip,
+            "limit": limit
         }
         message_list = await self._messages_api.v4_stream_sid_message_get(**params)
         return message_list.value
 
-    @retry
     async def blast_message(
```
symphony-youri, line 179: and if we want to push it one step further we could do the blast with the send message, allowing a list of stream ids
but given it is a different API call I would keep it a different service method

symphony/bdk/core/service/message/model.py
```diff
@@ -0,0 +1,104 @@
...
+
+        :return: the list of previews
+        """
+        return self._previews
+
+    @staticmethod
+    def _get_content(content):
+        if content is None:
+            raise MessageCreationError("Message content is mandatory")
+        if not content.startswith(MESSAGE_ML_START_TAG) and not content.endswith(MESSAGE_ML_END_TAG):
```
symphony-youri, line 81: we don't do it in Java but I was wondering if we should trim the string first in case you have something like:
`___<messageML>...`

symphony-youri (approved):

symphony/bdk/core/service/message/model.py
symphony-elias: In Java: https://github.com/finos/symphony-bdk-java/blob/main/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/model/Message.java#L175

symphony-elias: We'll keep things as is

PR #194, opened 2021-05-26
PLAT-10782: Support of multiple bot instances

symphony-elias:
### Ticket
PLAT-10782

### Description
Support of multiple bot instances by doing the following:

- using the `tag` field when listing or creating datafeeds (for DFv2 only)
- not updating the ack ID in case listener raises an `EventError` (for DFv2 only)
- logging a warning message when processing of events by listeners takes more than 30s (for DFv2 only)

### Checklist


symphony-youri (commented):
You can perhaps reuse the InjectorBot https://github.com/finos/symphony-bdk-java/blob/e107cf42832fa4a4c82960d0eea1e3505c85cd71/symphony-bdk-examples/bdk-multi-instances-example/src/main/java/com/symphony/bdk/examples/df2/InjectorBot.java#L37 and write a reader bot in Python to test it

symphony/bdk/core/service/datafeed/datafeed_loop_v1.py
```diff
@@ -53,3 +49,28 @@ async def recreate_datafeed(self):
         datafeed_id = response.id
         self._datafeed_repository.write(datafeed_id)
         self._datafeed_id = datafeed_id
+
+    async def _run_loop_iteration(self):
+        event_list = await self._read_datafeed()
+        done = await self._run_listener_tasks(event_list)
```
symphony-youri: done -> done_tasks? (just to suggest that it is a list)

symphony/bdk/core/service/datafeed/datafeed_loop_v1.py
```diff
@@ -53,3 +49,28 @@ async def recreate_datafeed(self):
         datafeed_id = response.id
         self._datafeed_repository.write(datafeed_id)
         self._datafeed_id = datafeed_id
+
+    async def _run_loop_iteration(self):
+        event_list = await self._read_datafeed()
```
symphony-youri: event_list -> events?

symphony-youri (commented):
Does the logging we implement with the context vars still work?

symphony-youri (commented):
LGTM, let's test it a bit manually/with examples so we can see for real that it works ^^ (not so easy to automate anyway)

symphony-elias:
> Does the logging we implement with the context vars still work?

Yes it works

examples/multiple_instances/reader.py
```diff
@@ -0,0 +1,76 @@
...
+        finally:
+            lock.unlock()
+
+    async def _reply_to_message(self, event):
+        message = event.message
+        text_message = get_text_content_from_message(message)
+        message_number = text_message.split(" ")[0]
+        logging.debug("Received message number %s", message_number)
+
+        processed_events = self._client.get_map("processed_events").blocking()
```
symphony-youri, line 44: can it be a problem to block from an async function?

docsrc/markdown/datafeed.md
```diff
@@ -113,4 +120,56 @@ datafeed:
     initialIntervalMillis: 500
     multiplier: 2
     maxIntervalMillis: 300000
-```
\ No newline at end of file
+```
+
+## Best practices
```
symphony-youri, line 125: I wonder how we should avoid duplicating this in the future?
I guess we should put this kind of documentation only on the developers website? (at least it could be shared but less open to contribution...)

symphony-youri (approved):
Disclaimer: I have not tested it locally, just reviewed the diffs

examples/multiple_instances/reader.py
symphony-elias: Indeed it is blocking but asynchronous maps in the hazelcast lib does not used asyncio and standard Python future objects but callbacks (see https://hazelcast.readthedocs.io/en/stable/api/proxy/map.html) and hazelcast.future.Future. This would make the implementation harder and less readable IMO.

PR #195, opened 2021-05-31
PLAT-10893 Check for duplicate slash commands

symphony-mariacristina:
### Ticket
[PLAT-10893](https://perzoinc.atlassian.net/browse/PLAT-10893)

### Description
Goal of this implementation is to check when registering a new slash command that is not equal to any of the already existing ones. A slash command is considered equal when both name value and mention_bot boolean are the same.

### Dependencies

### Checklist


symphony/bdk/core/activity/command.py
```diff
@@ -81,3 +80,12 @@ def matches(self, context: CommandContext) -> bool:
 
     async def on_activity(self, context: CommandContext):
         await self._callback(context)
+
+    def __eq__(self, o: object) -> bool:
+        if isinstance(o, SlashCommandActivity):
```
symphony-soufiane, line 84: do we need to check if "o" is None ?

symphony-soufiane (approved):
LGTM, just left one comment to avoid errors when comparing None object to a command.

symphony/bdk/core/activity/command.py
symphony-elias: No, if `o` is `None`, it is of type `NoneType` and `isinstance(o, SlashCommandActivity)` will return `False`.
Tested on the Python console:
Screenshot 2021-06-01 at 09 32 04
.

symphony/bdk/core/activity/registry.py
```diff
@@ -47,10 +47,18 @@ def register(self, activity: AbstractActivity):
         :param activity: any object inheriting from base :class:`AbstractActivity`
         """
         logger.debug("Registering new activity %s", activity)
+        self._pre_process_activity(activity)
         self._activity_list.append(activity)
 
+    def _pre_process_activity(self, activity: AbstractActivity):
+        if isinstance(activity, SlashCommandActivity):
```
symphony-elias: I suppose this is useless since it will call the __eq__ function of the correct class (cf. polymorphism).

symphony/bdk/core/activity/registry.py
```diff
@@ -47,10 +47,18 @@ def register(self, activity: AbstractActivity):
...
+            for act in self._activity_list:
+                if act.__eq__(activity):
```
symphony-elias: Pythonic way to call __eq__ is simply to write `activity == act`

symphony/bdk/core/activity/registry.py
symphony-mariacristina: Removed in the next commit

symphony/bdk/core/activity/registry.py
symphony-mariacristina: changed!

tests/core/activity/registry_test.py
```diff
@@ -222,3 +222,40 @@ async def test_slash_command(activity_registry, message_sent):
...
     assert slash_activity._callback == listener
+
+
+@pytest.mark.asyncio
+async def test_slash_same_command_name_and_mention(activity_registry, message_sent):
+    listener = AsyncMock()
+    mention_bot = True
+    command_name = "/command"
+
+    activity_registry.slash(command_name, mention_bot)(listener)
```
symphony-elias: Please consider having two different listener instances to check we do not blindly make a reference equality check

symphony-elias (changes requested):

tests/core/activity/registry_test.py
symphony-mariacristina: you are right, I missed that. Done in https://github.com/SymphonyPlatformSolutions/symphony-api-client-python/pull/195/commits/cd17c57c6512282cc8ad25df885042bc2a287681

symphony-elias (approved):

PR #197, opened 2021-06-01
PLAT-10895-Implement built-in help command

symphony-mariacristina:
### Ticket
[PLAT-10895](https://perzoinc.atlassian.net/browse/PLAT-10895)

### Description
Goal of this PR is to provide the implementation of a build-in help command with lists all the activity registered in the ActivityRegistry.
For each command found the following format will be used " - command_name - description (mention required/mention not required)" (example "/hello - command to say hello (mention required)")
This help command can also be overriden by manually implementing a new SlashCommandActivity, the code will make sure that any duplicates will be removed.

### Checklist

### DIFF
--- a/symphony/bdk/core/activity/command.py
+++ b/symphony/bdk/core/activity/command.py
@@ -94,7 +94,3 @@ class SlashCommandActivity(CommandActivity):
         if isinstance(o, SlashCommandActivity):
             return self._name == o._name and self._requires_mention_bot == o._requires_mention_bot
         return False
-
-    def __hash__(self) -> int:
-        return super().__hash__()
-
--- a/symphony/bdk/core/activity/command.py
+++ b/symphony/bdk/core/activity/command.py
@@ -63,7 +63,7 @@ class SlashCommandActivity(CommandActivity):
       - "{name}" otherwise
     """
 
-    def __init__(self, name, requires_mention_bot, callback):
+    def __init__(self, name, requires_mention_bot, callback, description=""):
         """
         :param name: the command name
         :param requires_mention_bot: if the command requires the bot mention to trigger the slash command
@@ -72,6 +72,16 @@ class SlashCommandActivity(CommandActivity):
         self._name = name
         self._requires_mention_bot = requires_mention_bot
         self._callback = callback
+        self._description = description
+
+    @property
+    def name(self) -> str:
+        return self._name
+
+    def build_command_description(self) -> str:
+        if self._requires_mention_bot:
+            return self._description + " (mention required)"
+        return self._description + " (mention not required)"
 
     def matches(self, context: CommandContext) -> bool:
         pattern = rf"@{context.bot_display_name} {self._name}" if self._requires_mention_bot else rf"{self._name}"
--- a/symphony/bdk/core/activity/command.py
+++ b/symphony/bdk/core/activity/command.py
@@ -79,9 +79,8 @@ class SlashCommandActivity(CommandActivity):
         return self._name
 
     def build_command_description(self) -> str:
-        if self._requires_mention_bot:
-            return self._description + " (mention required)"
-        return self._description + " (mention not required)"
+        return self._description + " (mention required)" if self._requires_mention_bot \
+            else self._description + " (mention not required)"
 
     def matches(self, context: CommandContext) -> bool:
         pattern = rf"@{context.bot_display_name} {self._name}" if self._requires_mention_bot else rf"{self._name}"
--- a/symphony/bdk/core/activity/help_command.py
+++ b/symphony/bdk/core/activity/help_command.py
@@ -1,29 +1,22 @@
 from symphony.bdk.core.activity.command import CommandContext, SlashCommandActivity
-from symphony.bdk.core.activity.registry import ActivityRegistry
-from symphony.bdk.core.service.message.message_service import MessageService
 from symphony.bdk.core.service.message.model import Message
+from symphony.bdk.core.symphony_bdk import SymphonyBdk
 
 
 class HelpCommand(SlashCommandActivity):
     """The help command is a particular CommandActivity which returns the list of all commands available for the
     specific bot
     The Slash command is triggered if we receive a MESSAGESENT event with message text:
-      - "@{bot_display name} {name}"
+      - "@{bot_display name} /help"
     """
 
-    def __init__(self, registry: ActivityRegistry, callback: MessageService):
-        super().__init__("/help", True, callback, "List available commands")
-        self._registry = registry
-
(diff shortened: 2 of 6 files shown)


symphony/bdk/core/activity/help_command.py
```diff
@@ -0,0 +1,29 @@
...
+    specific bot
+    The Slash command is triggered if we receive a MESSAGESENT event with message text:
+      - "@{bot_display name} {name}"
+    """
+
+    def __init__(self, registry: ActivityRegistry, callback: MessageService):
+        super().__init__("/help", True, callback, "List available commands")
+        self._registry = registry
+
+    def matches(self, context: CommandContext) -> bool:
```
symphony-elias: This is already implemented in the `SlashCommandActivity` class so this can be removed.

symphony/bdk/core/activity/help_command.py
```diff
@@ -0,0 +1,29 @@
...
+        pattern = rf"@{context.bot_display_name} {self._name}"
+        return pattern == context.text_content
+
+    async def on_activity(self, context: CommandContext):
+        activity_list = self._registry.activity_list
+        help_message = ""
+        for act in activity_list:
+            if not isinstance(act, HelpCommand):
```
symphony-elias: The description of the help command can be added to the output as well (e.g. man pages do show the options to display the help command)

symphony/bdk/core/activity/help_command.py
```diff
@@ -0,0 +1,29 @@
...
+                help_message += "<li>" + act.name + " - " + act.build_command_description() + "</li>"
+        if not activity_list:
```
symphony-elias: It means there are cases where we don't send the `help_message` ? I would be in favor of always sending a message.

symphony/bdk/core/activity/help_command.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class HelpCommand(SlashCommandActivity):
+    """The help command is a particular CommandActivity which returns the list of all commands available for the
+    specific bot
+    The Slash command is triggered if we receive a MESSAGESENT event with message text:
+      - "@{bot_display name} {name}"
+    """
+
+    def __init__(self, registry: ActivityRegistry, callback: MessageService):
```
symphony-elias: IMO we could simplify the usage of the `HelpCommand` by having to pass the `SymphonyBdk` object only.

symphony/bdk/core/activity/help_command.py
```diff
@@ -0,0 +1,29 @@
...
+from symphony.bdk.core.activity.registry import ActivityRegistry
+from symphony.bdk.core.service.message.message_service import MessageService
+from symphony.bdk.core.service.message.model import Message
+
+
+class HelpCommand(SlashCommandActivity):
+    """The help command is a particular CommandActivity which returns the list of all commands available for the
+    specific bot
+    The Slash command is triggered if we receive a MESSAGESENT event with message text:
+      - "@{bot_display name} {name}"
```
symphony-elias: `"@{bot_display name} /help"` ?

symphony/bdk/core/activity/help_command.py
symphony-mariacristina: I will add /help as well in the returning list of commands

symphony/bdk/core/activity/help_command.py
symphony-mariacristina: I will add a message to say that there are no commands available

symphony-mariacristina: Well actually if we include the /help we will always have at least that one returned

symphony/bdk/core/activity/command.py
```diff
@@ -72,6 +72,16 @@ def __init__(self, name, requires_mention_bot, callback):
...
         self._requires_mention_bot = requires_mention_bot
         self._callback = callback
+        self._description = description
+
+    @property
+    def name(self) -> str:
+        return self._name
+
+    def build_command_description(self) -> str:
+        if self._requires_mention_bot:
```
symphony-elias: This could even be a one-liner: 
`return self._description + " (mention required)" if self._requires_mention_bot else self._description + " (mention not required)"`
But this way is fine, too.

symphony/bdk/core/activity/help_command.py
```diff
@@ -0,0 +1,22 @@
...
+
+class HelpCommand(SlashCommandActivity):
+    """The help command is a particular CommandActivity which returns the list of all commands available for the
+    specific bot
+    The Slash command is triggered if we receive a MESSAGESENT event with message text:
+      - "@{bot_display name} /help"
+    """
+
+    def __init__(self, bdk: SymphonyBdk):
+        super().__init__("/help", True, bdk.messages(), "List available commands")
```
symphony-elias: IMO putting the `MessageService` instance as callback is misleading as this was designed to be an async function accepting a `CommandContext`, which will then be called when the message matches the parameters (command name and mention bot). I'd prefer to have:

- either None as callback and keep the overriden `on_activity` method
- or create a specific method and put it as callback in `super().__init__`

tests/core/activity/help_command_test.py
```diff
@@ -0,0 +1,101 @@
...
+
+    activity_registry.slash("/hello")(listener)
+    activity_registry.register(HelpCommand(bdk))
+
+    assert len(activity_registry.activity_list) == 2
+    help_activity = activity_registry.activity_list[1]
+    assert isinstance(help_activity, HelpCommand)
+    assert help_activity.matches(command_context)
+    await help_activity.on_activity(command_context)
+    help_activity._callback.send_message.assert_called_once()
```
symphony-elias: We should check the message sent IMHO

tests/core/activity/help_command_test.py
```diff
@@ -0,0 +1,101 @@
...
+@pytest.mark.asyncio
+async def test_help_command_no_other_commands_found(bdk, activity_registry, command_context):
+    activity_registry.register(HelpCommand(bdk))
+
+    assert len(activity_registry.activity_list) == 1
+    help_activity = activity_registry.activity_list[0]
+    assert isinstance(help_activity, HelpCommand)
+    assert help_activity.matches(command_context)
+    await help_activity.on_activity(command_context)
+    help_activity._callback.send_message.assert_called_once()
```
symphony-elias: Same here, we should check for the sent message

symphony/bdk/core/activity/help_command.py
symphony-mariacristina: I will change it to pass None as callback and I will call directly the bdk object inside the on_activity function

symphony-elias (approved):
LGTM !

PR #199, opened 2021-06-01
PLAT-10868: Remove certificates and private keys credentials

symphony-youness:
### Ticket
[PLAT-10868](https://perzoinc.atlassian.net/browse/PLAT-10868)

### Description
We want to remove all hardcoded credentials from the repo.

To do so, we use pytest fixture to generate these when running tests. These fixtures are **session scoped** so they will generate credentials (rsa key, certificate) only the first time they are requested.

These fixtures are put in a separate `conftest.py` to be accessed by all tests in the suite.

- remove credentials
- Fix related tests

### Checklist


tests/conftest.py
```diff
@@ -0,0 +1,40 @@
...
+@pytest.fixture(name="rsa_key", scope="session")
+def fixture_rsa_key(root_key):
+    return root_key.private_bytes(
+        encoding=serialization.Encoding.PEM,
+        format=serialization.PrivateFormat.PKCS8,
+        encryption_algorithm=serialization.NoEncryption()).decode("utf-8")
+
+
+@pytest.fixture(name="certificate", scope="session")
+def fixture_certificate(root_key):
```
symphony-youness, line 29: If I understood correctly, we don't actually need to remove .cert files but I thought I could add that too to keep credential usage aligned.

PR #234, opened 2021-10-06
PLAT-10714: Update Room V3 API to Pin/Unpin a message

symphony-youness:
### Ticket
[PLAT-10714](https://perzoinc.atlassian.net/browse/PLAT-10714)

### Description
The goal is to update the generated code in the python BDK to add the new parameter pinnedMessageId


symphony-elias (approved):
LGTM !

PR #241, opened 2021-11-10
PLAT 11396: Support update message endpoints

symphony-youness:
[PLAT 11396](https://perzoinc.atlassian.net/browse/PLAT-11396)

Added new message update endpoint to `message_service` class
- Mentioned in markdown doc 
- Added message update example in the example folder

### DIFF
--- a/symphony/bdk/core/service/message/message_service.py
+++ b/symphony/bdk/core/service/message/message_service.py
@@ -464,6 +464,36 @@ class MessageService(OboMessageService):
 
         return offset_based_pagination(search_messages_one_page, chunk_size, max_number)
 
+    @retry
+    async def update_message(self, stream_id: str, message_id: str, message: Union[str, Message], data=None,
+                             version: str = "") -> V4Message:
+        """Update an existing message. The existing message must be a valid social message, that has not been deleted.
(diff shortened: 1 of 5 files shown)


symphony/bdk/core/service/message/message_service.py
```diff
@@ -464,6 +464,36 @@ async def search_messages_one_page(skip, limit):
...
+        message_object = message if isinstance(message, Message) else Message(content=message, data=data, version=version)
+
+        params = {
+            "sid": stream_id,
+            "mid": message_id,
+            "session_token": await self._auth_session.session_token,
+            "key_manager_token": await self._auth_session.key_manager_token,
+            "message": message_object.content,
+            "data": message_object.data,
+            "version": message_object.version
```
symphony-youri, line 493: uh, I just discovered this parameter... I think I've never seen it used 🙄

symphony-youri (approved):

symphony-mariacristina (approved):

symphony-elias (approved):

symphony/bdk/core/service/message/message_service.py
symphony-youness: Same here, I saw it was specified in the docs and sendMessage method so I used it.

PR #246, opened 2021-11-29
Upgrade dependencies

symphony-youri:
Mainly for aiohttp to address https://security.snyk.io/vuln/SNYK-PYTHON-AIOHTTP-1584144
(even though the BDK is not vulnerable to such issues as it does not
expose a web application)

Also upgraded tenacity with an API change.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1,21 +1,33 @@
 [[package]]
 name = "aiohttp"
-version = "3.7.4.post0"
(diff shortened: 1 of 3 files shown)


symphony-elias (approved):

symphony-youri:
As the BDK is not really vulnerable let's ship this in the next release but I would say we don't need to make a specific one.

Issue #251, opened 2022-01-05
Module system


symphony-elias:
The purpose of this issue is to introduce and specify a new module system for the BDK library as in Java: https://github.com/finos/symphony-bdk-java/issues/479

We want to empower contributors to create their own modules or plugins in separate repositories, without necessary having to make their contributions directly in the BDK codebase.

We could think of something like:
```
async def run():
    async with SymphonyBdk(BdkConfigLoader.load_from_symphony_dir("config.yaml")) as bdk:
        extension = bdk.extension(MyExtension)
```
with the `extension` having access to the BDK configuration, the session/km tokens, etc.

PR #261, opened 2022-02-16
Extension API PoC

symphony-elias:
### Ticket
#251

### Description
PoC of what could be a module system similar to the one implemented in [symphony-bdk-java](https://github.com/finos/symphony-bdk-java/blob/main/docs/extension.md)

### DIFF
--- a/examples/services/extension_example.py
+++ b/examples/services/extension_example.py
@@ -5,13 +5,13 @@ from pathlib import Path
 from symphony.bdk.core.auth.auth_session import AuthSession
 from symphony.bdk.core.config.loader import BdkConfigLoader
 from symphony.bdk.core.config.model.bdk_config import BdkConfig
-from symphony.bdk.core.extension import BdkRetryConfigAware, retry_aware
-from symphony.bdk.core.symphony_bdk import SymphonyBdk, register_extension
+from symphony.bdk.core.extension import retry_aware, register_extension
+from symphony.bdk.core.symphony_bdk import SymphonyBdk
 
 
 @register_extension
 @retry_aware  # so that _retry_config attribute is set and @retry decorator can be used with the same retry config
-class MyExtension(BdkRetryConfigAware):
+class MyExtension:
     def __init__(self):
         self._bot_session = None
         self._config = None
--- a/examples/services/extension_example.py
+++ b/examples/services/extension_example.py
@@ -3,6 +3,7 @@ import logging.config
 from pathlib import Path
 
 from symphony.bdk.core.auth.auth_session import AuthSession
+from symphony.bdk.core.client.api_client_factory import ApiClientFactory
 from symphony.bdk.core.config.loader import BdkConfigLoader
 from symphony.bdk.core.config.model.bdk_config import BdkConfig
 from symphony.bdk.core.extension import retry_aware, register_extension
@@ -15,6 +16,7 @@ class MyExtension:
     def __init__(self):
         self._bot_session = None
         self._config = None
+        self._api_client_factory = None
 
     def set_auth_session(self, auth_session: AuthSession):
         logger.debug("Registering auth session")
@@ -23,6 +25,10 @@ class MyExtension:
     def set_configuration(self, bdk_config: BdkConfig):
         logger.debug("Registering config: " + bdk_config.host)
         self._config = bdk_config
+
+    def set_api_client_factory(self, api_client_factory: ApiClientFactory):
+        logger.debug("Registering api client factory: " + str(api_client_factory))
+        self._api_client_factory = api_client_factory
 
     def get_service(self):
         return self
--- a/examples/services/extension_example.py
+++ b/examples/services/extension_example.py
@@ -1,0 +1,41 @@
+import asyncio
+import logging.config
+from pathlib import Path
+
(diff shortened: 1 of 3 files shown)


symphony/bdk/core/extension.py
```diff
@@ -0,0 +1,54 @@
+import logging
+from abc import ABC, abstractmethod
+
+from symphony.bdk.core.auth.auth_session import AuthSession
+from symphony.bdk.core.config.model.bdk_config import BdkConfig
+
+logger = logging.getLogger(__name__)
+
+
+class BdkConfigAware(ABC):
```
symphony-elias, line 24: Event if we use duck typing, I think it is nice to have abstract classes to document in the code which interface we expect.

symphony/bdk/core/extension.py
```diff
@@ -0,0 +1,54 @@
...
+            logging.debug("Extension is not a authentication aware")
+        except TypeError:
+            logging.warning("set_auth_session method must have a single positional argument")
+
+    def get_service(self, extension_type):
+        extension = self._extensions[extension_type]
+        if extension is None:
+            raise ValueError(f"Extension {extension_type} not defined")
+
+        return extension.get_service()  # will raise an AttributeError or TypeError if get_service() not defined
```
symphony-elias, line 88: In the real implementation, we could wrap the errors into a dedicated error type.

symphony/bdk/core/extension.py
symphony-elias: ⚠️ For the retry mechanism, we use `@retry` decorators on methods that needs to be retried. This decorator checks retry configuration in field `self._retry_config` assuming such field exists. We will have to properly document that.
`@retry_aware` sets the attribute `_retry_config` with the correct value, in case class is an extension

examples/services/extension_example.py
```diff
@@ -0,0 +1,41 @@
...
+from pathlib import Path
+
+from symphony.bdk.core.auth.auth_session import AuthSession
+from symphony.bdk.core.config.loader import BdkConfigLoader
+from symphony.bdk.core.config.model.bdk_config import BdkConfig
+from symphony.bdk.core.symphony_bdk import SymphonyBdk
+
+
+class MyExtension:
+    def __init__(self):
```
symphony-youri, line 16: how do we get access to the ApiClientFactory/pod client?

examples/services/extension_example.py
```diff
@@ -0,0 +1,41 @@
...
+        return self
+
+    async def get_session_token(self):
+        return await self._bot_session.session_token
+
+
+async def run():
+    async with SymphonyBdk(BdkConfigLoader.load_from_symphony_dir("config.yaml")) as bdk:
+        bdk.extensions().register(MyExtension)
+        service = bdk.extensions().get_service(MyExtension)
```
symphony-youri, line 42: you might even wonder, do we need to register and get the service? vs just getting the service

examples/services/extension_example.py
symphony-elias: I didn't implement that on purpose because I wanted to just agree on the overall design.
To access the pod client, we will inject the `ApiClientFactory` object in `symphony/bdk/core/symphony_bdk.py` (line 121) when creating the `ExtensionService`.

examples/services/extension_example.py
symphony-elias: Yes, I remember we discussed that during the design of the extension API in the Java BDK.
* We could have extension that don't want to expose services (like background tasks)
* I guess we want to stay close to the Java BDK API to not confuse customers
* We could provide both ways: programmatic registration to stay close to the Java BDK and a decorator like `@extension.register` to register an extension

examples/services/extension_example.py
```diff
@@ -0,0 +1,43 @@
...
+
+from symphony.bdk.core.auth.auth_session import AuthSession
+from symphony.bdk.core.config.loader import BdkConfigLoader
+from symphony.bdk.core.config.model.bdk_config import BdkConfig
+from symphony.bdk.core.extension import BdkRetryConfigAware, retry_aware
+from symphony.bdk.core.symphony_bdk import SymphonyBdk, register_extension
+
+
+@register_extension
+@retry_aware  # so that _retry_config attribute is set and @retry decorator can be used with the same retry config
```
symphony-elias, line 14: * `@retry` decorator calls `getattr(self, '_retry_config')` here: https://github.com/finos/symphony-bdk-python/blob/main/symphony/bdk/core/retry/__init__.py#L32
* we cannot decouple `@retry_aware` from `@register_extension` because the `@register_extension` actually instantiates the object.

examples/services/extension_example.py
symphony-elias: Implemented now

examples/services/extension_example.py
symphony-elias: An open question: do we want to give the ability to register extensions by specifying a package name (as a string)? Do we want to give the ability to get a service by specifying a string, like `"my.package.MyExtension"`? 
If we go for that, implementation will be slightly different as we will have to dynamically import the extensions using importlib.

symphony-youri: given you have the extension in your project at that stage, is there a value in doing it dynamically?

PR #265, opened 2022-03-01
Using latest openapi generator + generating profile-manager api

symphony-elias:
### Ticket
PLAT-11835

### Description
In this implementation we:
- upgraded the openapi-generator version to use 5.4.0. Custom branch used for the jar generation can be found [here](https://github.com/OpenAPITools/openapi-generator/pull/11758).
- custom templates that were part of the templates module were integrated in the custom jar, hence removed from this project
- all existing APIs and models were regenerated using the new openapi-generator version
- as prerequisite for the group extension (coming soon), Group APIs were generated

### Checklist


symphony-youri:
Why isn't https://github.com/OpenAPITools/openapi-generator/pull/11758 something that could be merged?

We could use a better description of the changes, it seems a lot was done here (changing the templates, the generator, adding a new swagger spec)

Would there be an interest in splitting the groups separated (an entire separated folder tree with its own generated code folder), so that if we decide to split it we just have to move the entire folder away? (it seems a bit YAGNI though)
Or maybe because the update cycle of the group spec vs the pod/agent spec will be different? we will update one or not the other (but that we can deal with with fixed commit ids)

symphony-youri (approved):

PR #268, opened 2022-03-17
Extension mechanism

symphony-mariacristina:
### Description
The Extension API is available through the module `symphony.bdk.core.extension`. Extensions are registered _programmatically_ via the `ExtensionService`. A _Service Provider_ extension is a specific type of extension loaded on demand when calling the `ExtensionService#service(MyExtensionType)` method.

To make your extension _Service Provider_, your extension definition must implement the method `get_service(self)`.
```python
# Usage example.
async def run():
        extension_service = bdk.extensions()
        extension_service.register(MyBdkExtensionService)
        service = extension_service.service(MyBdkExtensionService)
        service.say_hello("Symphony")
        
asyncio.run(run())
```
#### BDK Aware Extensions
The BDK Extension Model allows extensions to access to some core objects such as the configuration or the api clients.
Developers that wish to use these objects are free to implement a set of abstract base classes all suffixed with the `Aware` keyword.
List of Aware classes:
- `BdkConfigAware`-  allows extensions to read the `BdkConfig`: 
- `BdkApiClientFactoryAware` - can be used by extensions that need to use the `ApiClientFactory`
- `BdkAuthenticationAware`- can be used by extensions that need to rely on the service account authentication session (`AuthSession`)
### Symphony Groups Extension
As part of this PR we also leveraged this new **Extension Mechanism** to implement a new custom extension to support the [Symphony Groups APIs](https://developers.symphony.com/restapi/reference/addmembertogroup).

### DIFF
--- a/symphony/bdk/ext/group.py
+++ b/symphony/bdk/ext/group.py
@@ -22,26 +22,11 @@ async def refresh_bearer_token_if_unauthorized(retry_state):
     :param retry_state: current retry state
     :return: True if we want to retry, False otherwise
     """
-    service_oauth_session = retry_state.args[0]._oauth_session
-    return await _refresh_if_unauthorized(retry_state, service_oauth_session)
-
-
-async def refresh_session_token_if_unauthorized(retry_state):
-    """Function used by the retry decorator to refresh the session token if conditions apply
-
-    :param retry_state: current retry state
-    :return: True if we want to retry, False otherwise
-    """
-    service_auth_session = retry_state.args[0]._session
-    return await _refresh_if_unauthorized(retry_state, service_auth_session)
-
-
-async def _refresh_if_unauthorized(retry_state, service_session):
     if retry_state.outcome.failed:
         exception = retry_state.outcome.exception()
         if is_network_or_minor_error(exception):
             if is_unauthorized(exception):
-                await service_session.refresh()
+                await retry_state.args[0]._oauth_session.refresh()
             return True
     return False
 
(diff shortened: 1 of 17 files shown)


symphony/bdk/ext/group.py
```diff
@@ -0,0 +1,149 @@
...
+        client = self._api_client_factory.get_client("/profile-manager")
+        # to enable setting the authorization header defined in api_client.ApiClient.update_params_for_auth
+        client.configuration.auth_settings = lambda: {
+            "bearerAuth": {"in": "header", "type": "bearer", "key": "Authorization",
+                           "value": "Bearer " + self._oauth_session.bearer_token()}}
+        # needed for the x_symphony_host parameter to allow empty value
+        client.configuration._disabled_client_side_validations = ["minLength"]
+        self._group_api = GroupApi(client)
+
+    async def insert_group(self, create_group: CreateGroup) -> ReadGroup:
```
symphony-youri, line 75: I was going to say we should name it create group to be consistent with other services but then in the Java BDK it is named like that, I missed it during the Java BDK review :(

symphony/bdk/ext/group.py
```diff
@@ -0,0 +1,149 @@
...
+
+    async def add_member_to_group(self, group_id: str, user_id: int) -> ReadGroup:
+        """Add a new user to an existing group.
+        See: `Add a new user to an existing group <https://developers.symphony.com/restapi/reference/addmembertogroup>`_
+
+        :param group_id: the ID of the group in which to add the user
+        :param user_id: The ID of the user to be added into the group
+        :return: the updated group
+        """
+        member = Member(member_id=user_id, member_tenant=extract_tenant_id(user_id))
```
symphony-youri, line 156: Mmm I missed that in the Java BDK we were doing that (guessing the tenant id from the user id) (which the profile manager API should do for us already...)
So I should probably change the WDK implem too (although we only do it partially for this method and not for the creation)

symphony-elias:
@symphony-mariacristina should we document how to leverage the retry mechanism in extensions? I would say so because it differs much from the Java BDK

symphony-elias (approved):
LGTM

symphony-elias:
Implements #251

PR #287, opened 2022-05-25
Add silent message update support

yinan-symphony:
- Update Symphony APIs spec version

 - Add silent flag support according to the new spec.
   When silent is true, the new updated message is marked as read, otherwise is unread.
   True is the default value.

### Description
Closes #[ISSUE NUMBER]

### Dependencies

### Checklist


api_client_generation/generate.sh
```diff
@@ -4,7 +4,7 @@ code_gen_dir=`pwd`
 project_root=$code_gen_dir/..
 echo $code_gen_dir
 
-commit_hash=136b530512eea112de73591906051871d034750a
```
symphony-soufiane, line 7: When we update this commit hash, we might want to run the sh script to update the generated code. See here https://github.com/finos/symphony-bdk-python/blob/main/docsrc/markdown/tech/generated_api.md

symphony-soufiane: And include the new generated files in the PR of course.

symphony/bdk/core/service/message/message_service.py
```diff
@@ -466,7 +466,7 @@ async def search_messages_one_page(skip, limit):
 
     @retry
     async def update_message(self, stream_id: str, message_id: str, message: Union[str, Message], data=None,
-                             version: str = "") -> V4Message:
+                             silent=True, version: str = "") -> V4Message:
```
symphony-soufiane: I'm wondering if adding this parameter doesn't break the backward compatibility. I would rather add a new method that takes this new argument, and I call it in this one with the default silent value (True).

yinan-symphony: i checked this point actually. this `silent=True` syntax in python is working as an optional arg with the default value. so user can choose to specify it or simply ignore it, in this case the default value is applied.
you can see that the unit test does not change the signature, but it is still working.

api_client_generation/generate.sh
yinan-symphony: sure, i am doing it now

yinan-symphony: @symphony-soufiane done.

symphony-soufiane (approved):

PR #299, opened 2022-06-30
Bump main branch version to 2.4.dev0

symphony-soufiane:
Bump main branch version to 2.4.dev0

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "symphony_bdk_python"
-version = "2.3.dev0"
+version = "2.4.dev0"
 license = "Apache-2.0"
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "symphony_bdk_python"
-version = "2.3.dev0"
+version = "2.4.dev0"
```
yinan-symphony: since 2.4 is ready to release, should we not use 2.5.dev0 instead ?

symphony-soufiane: You're right. I don't remember why we kept using the same version as the release in main, with "dev" suffix. I corrected it in the last commit.

yinan-symphony (approved):

PR #302, opened 2022-07-04
#295 Bug fix: presence not working with DF2

### LINKED ISSUE #295: Online Presence not working in Python BDK
## Bug Report
Online presence not working
 
### Steps to Reproduce:
1. Create a Python Demo Bot using the Python BDK
2. Launch bot, authenticate and login
3. Send/Receive Messages to the bot, but presence does not show as online.

### Expected Result:
The Bots online presence should have a green online status.

1654866691699_devCertBot2136_online status

### Actual Result:
Starting up the demo bot, runs fine. It looks good and replies with the @mention followed by /command but the bot's online status at the Symphony App is off line with a grey box.

### Environment:
develop2.symphony.com

symphony-soufiane:
### Description
Closes #295 
When using DF2, the bot presence is handled in Agent side. The datafeed ackId has been previously updated to include a boolean field update_presence in addition to the ackId string. This boolean field allows Agent to update bot presence along with the DF heartbeat. The field default value is True and cannot be changed by the bot (it is only customisable via API calls).

The issue preventing the presence update was happening because the default value of update_presence, which is True, was generated as None in BDK Python, which in turn is mapped to False in Agent side. Consequently, Agent was skipping bot presence update.

In this commit, we updated the generated code to use the latest that we changed to set the default value for optiona vars, if provided, otherwise None is used. Previously, None was always set in any case.

### Checklist
- [x] Referenced an issue in the PR title or description

### DIFF
--- a/symphony/bdk/gen/agent_model/ack_id.py
+++ b/symphony/bdk/gen/agent_model/ack_id.py
@@ -247,7 +247,7 @@ class AckId(ModelNormal):
         self._visited_composed_classes = _visited_composed_classes + (self.__class__,)
 
         self.ack_id: str = None
-        self.update_presence: bool = None
+        self.update_presence: bool = True
         for var_name, var_value in kwargs.items():
             if var_name not in self.attribute_map and \
                         self._configuration is not None and \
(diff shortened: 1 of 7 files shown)


symphony-soufiane:
The fix in openapi-generator fork's branch that is used to package the openapi-generator-cli.jar that we added in this PR, can be found here: https://github.com/symphony-soufiane/openapi-generator/commit/f0667fb31689ff4a495f9411573e069568489cdd

yinan-symphony (approved):

PR #311, opened 2022-10-31
Fix Datahose loop

symphony-soufiane:
### Description
In the new update of Datahose api, the field filters has been renamed to eventTypes. In this commit, we apply this update in BDK Python. 
The generated code has been updated as well according to the last changes in symphony-api-specs.


pyproject.toml
```diff
@@ -42,7 +42,7 @@ requires = ["poetry-core>=1.0.0"]
 build-backend = "poetry.core.masonry.api"
 
 [tool.pytest.ini_options]
-addopts = "--junitxml=test-results/junit.xml --cov --cov-report=html"
+#addopts = "--junitxml=test-results/junit.xml --cov --cov-report=html"
```
yinan-symphony: is it on purpose?

examples/datafeed/datahose.py
```diff
@@ -6,13 +6,15 @@
...
+from symphony.bdk.gen.agent_model.v4_instant_message_created import V4InstantMessageCreated
+from symphony.bdk.gen.agent_model.v4_message_suppressed import V4MessageSuppressed
 from symphony.bdk.gen.agent_model.v4_room_updated import V4RoomUpdated
 from symphony.bdk.gen.agent_model.v4_room_created import V4RoomCreated
 from symphony.bdk.gen.agent_model.v4_message_sent import V4MessageSent
 
 
 async def run():
-    config = BdkConfigLoader.load_from_symphony_dir("config.yaml")
+    config = BdkConfigLoader.load_from_file("/Users/soufiane/.symphony/rsa-devx3/config.yaml")
```
yinan-symphony: I believe this is not meant to be commit ?

pyproject.toml
symphony-soufiane: Nope, my bad. This was to enable compilation on Pycharm. thanks!

yinan-symphony (approved):

PR #318, opened 2022-12-02
Feature/security scanning

yinan-symphony:
### Description
Closes #[ISSUE NUMBER]

### Dependencies

### Checklist


.github/workflows/cve-scanning-python.yml
```diff
@@ -0,0 +1,35 @@
+name: CVE Scanning for Python
+
+on:
+  pull_request:
+    paths:
+      - 'python/pyproject.toml'
+      - 'python/safety-policy.yml'
+      - '.github/workflows/cve-scanning-python.yml'
```
symphony-soufiane: where are these files? 
I don't see safety-policy.yml.
Pyproject.toml is not under /python. pyproject.toml but at root.

.github/workflows/cve-scanning-python.yml
```diff
@@ -0,0 +1,35 @@
...
+
+env:
+  PYTHON_KEYRING_BACKEND: keyring.backends.null.Keyring
+
+jobs:
+  scan:
+    name: Build and test App
```
symphony-soufiane, line 17: Do we need this in cve-scanning  ? We already have the build/test in https://github.com/finos/symphony-bdk-python/blob/main/.github/workflows/build.yml

.github/workflows/cve-scanning-python.yml
symphony-soufiane: Also do we want this to run only when those files are updated, or on every PR? Is it not better to run them on evert PR to check the packages versions more often?

requirements.txt
```diff
@@ -0,0 +1,21 @@
+aiohttp==3.8.3 ; python_version >= "3.8" and python_version < "4.0"
```
symphony-soufiane: for my curiosity, what is the purpose of this file please?

symphony-soufiane (commented):
left few comments :)

.github/workflows/cve-scanning-python.yml
yinan-symphony: @TheJuanAndOnly99 @maoo Could you please advise ?

requirements.txt
yinan-symphony: @TheJuanAndOnly99 @maoo please advise

.github/workflows/cve-scanning-python.yml
maoo: Yes, that is correct. Apologies for the mistake. Please adapt the file path as needed.

.github/workflows/cve-scanning-python.yml
maoo: Yes, i believe that it is necessary, since the scanning will occur on the libraries installed in the environment.

requirements.txt
maoo: If you use poetry, the requirements.txt should not be needed. I believe that Juan pushed it by mistake in the PR as it was not listed in .gitignore . Feel free to remove it.

yinan-symphony: just found it.  it is used by safety , https://github.com/pyupio/safety
@symphony-soufiane

.github/workflows/cve-scanning-python.yml
yinan-symphony: i will change the files' path accordingly.
this action is going to scan the dependencies only, and the dependencies will changed only when these files get changed, so no need to run the action for every PR, but only when those files got changed.

symphony-soufiane: ok so let's say we have a dependency that becomes outdated or vulnerable but we didn't change these files, what's the strategy to detect and correct that? Is there something else separated from this PR dedicated to this purpose?

yinan-symphony: hmm, right. I see what you mean. I agree with you.
However it is not clear what would be the exact behavior from their doc,
_When a pull request contains any files that are not matched by on:pull_request:paths-ignore or on:pull_request:paths, the workflow runs the actions and scans all of the files changed in the pull request_
I understood the scan is still going to run if any files not matched, we could merge and see what happen and update this config accordingly.

maoo: Good point @symphony-soufiane , I'd suggest to add - at the top of the actions - the following block, which will run the scan at the beginning/end of each working day:

```
on:
  schedule:
    - cron: '0 8,18 * * 1-5'
```

This would cover the case you describe. /cc @TheJuanAndOnly99

yinan-symphony: thx @maoo @symphony-soufiane

symphony-soufiane (approved):

PR #330, opened 2023-09-06
Expose Blast to OBO

anthony-symphony:
### Description

This is to expose the blast_message request to OBO as it has been recently added in Symphony.


yinan-symphony (approved):
LGTM

PR #339, opened 2024-03-19
Update cryptography aiohttp urllib3

anthony-symphony:
### Description
Update libs to fix vulnerabilities


yinan-symphony (approved):
LGTM

PR #345, opened 2024-10-28
Element datetime fix

### LINKED ISSUE #344: V4SymphonyElementAction automatically transform strings to datetime and other types
### Steps to Reproduce:

1. Create a messageML with a text-field.
2. Fill out the text-field with something that can be transformed to a datetime such as 20251023
3. Create and start Reading from Datafeed
4. Fill out form with 20251023 and submit

Alternatively:

1. Generate Python Bot with the Symphony Bot Generator
2. Send @mention /gif
3. Fill out form with 20251023 and submit.

### Expected Result:
The value should come in as a String and let the developer action against it as there is nothing in the response that indicates what type it should be. The type should not be inferred.

For the generated Python BDK:
Expected result is that it should return with "Category is 20251023"

### Actual Result:
The type is being inferred. Even though I put 20251023 and it can be inferred to be a valid date, in this case I am not expecting a datetime as it could be an ID or something else.

For the generated Python BDK after hitting submit you get "Category is 2025-10-23" which shows it being transformed to a datetime.

### Environment:
Python 3.11.1
Windows 10
symphony_bdk_python 2.7.1

### Additional Context:
Modifying Lines 90 and 263 of v4_symphony_elements_action.py to only include str and none_type fixes this issue but, may be considered a breaking change as there may developers currently expecting us to transform it to datetimes and other types.

https://github.com/finos/symphony-bdk-python/blob/e31a70140a5b873612454484c0ca7ffd2008cb23/symphony/bdk/gen/agent_model/v4_symphony_elements_action.py#L90

`'form_values': ({str: (str, none_type)}, none_type),  # noqa: E501`

https://github.com/finos/symphony-bdk-python/blob/e31a70140a5b873612454484c0ca7ffd2008cb23/symphony/bdk/gen/agent_model/v4_symphony_elements_action.py#L263

`self.form_values: Union[{str: (str, none_type)}] = None`

anthony-symphony:
### Description
Closes #344

Updated Code Generator to accept local files.
Added local file of the Symphony Agent Open API Spec with formValues modified


anthony-symphony:
security-checks failed due to https://data.safetycli.com/v/70612/f17/

Based on the note, this is a non-issue. snyk also did not flag this:

NOTE: The maintainer and multiple third parties believe that this vulnerability isn't valid because users shouldn't use untrusted templates without sandboxing.

yinan-symphony (approved):

PR #350, opened 2025-01-16
Added a new auth_token property to AuthSession class

broHeryk:
### Description
The reason for opening this PR is missing the auth_token property (which is present in[ java BDK](https://github.com/finos/symphony-bdk-java/blob/f0948e215a8ff9b558ad39732042e134172055fa/symphony-bdk-core/src/main/java/com/symphony/bdk/core/auth/impl/AuthSessionImpl.java#L64)). There's no other way except calling internal(private) methods so it's useful to expose this token so it's accessible. 

### Checklist


yinan-symphony:
@broHeryk since the repo has not been update since a while, some git actions are out of to date, pls update them.

yinan-symphony (approved):
👍  LGTM

PR #391, opened 2026-06-29
fix(datafeed): retry ClientPayloadError and reset _running so the loo…

Alex-Nalin:
…p can restart

A truncated datafeed read raises aiohttp.ClientPayloadError, which was not classified as a transient error, so read_datafeed_retry re-raised it and the datafeed loop crashed. The loop then could not be restarted because AbstractDatafeedLoop._run_loop left self._running = True on an abnormal exit (only stop() reset it), so DatafeedLoopV2.start() raised "The datafeed service V2 is already started" on every restart. The combined effect turned a transient network blip into a permanent, silent event-loss outage.

Changes:
- strategy.is_client_timeout_error now treats ClientPayloadError as transient (also benefits datahose and auth-refresh paths).
- AbstractDatafeedLoop._run_loop resets self._running = False in a finally, so the loop is restartable after any exit (stop, cancellation, or escaped exception). The concurrent-start guard in V1/V2 start() is unchanged.
- Add regression tests for both behaviours.

### Description
Closes #[ISSUE NUMBER]

### Dependencies

### Checklist

### DIFF
--- a/symphony/bdk/core/retry/strategy.py
+++ b/symphony/bdk/core/retry/strategy.py
@@ -1,6 +1,6 @@
 from asyncio import TimeoutError
 
-from aiohttp import ClientConnectionError
+from aiohttp import ClientConnectionError, ClientPayloadError
 from tenacity import RetryCallState
 
 from symphony.bdk.core.auth.exception import AuthUnauthorizedError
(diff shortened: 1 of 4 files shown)


Alex-Nalin:
Problem
A truncated datafeed read raises aiohttp.ClientPayloadError. This is not classified as a transient error, so read_datafeed_retry re-raises it and the datafeed loop crashes. The loop then cannot be restarted: AbstractDatafeedLoop._run_loop sets self._running = True and only stop() resets it, so after an abnormal exit _running stays True and DatafeedLoopV2.start() raises RuntimeError: The datafeed service V2 is already started on every restart attempt.

Net effect: a transient network blip becomes a permanent, silent event-loss outage. Observed in production as restart #1 (ClientPayloadError) followed by 500+ × "already started".

Changes
core/retry/strategy.py — is_client_timeout_error now treats ClientPayloadError as transient (also benefits the datahose and auth-refresh paths that share this predicate).
core/service/datafeed/abstract_datafeed_loop.py — _run_loop resets self._running = False in a finally, so the loop is restartable after any exit (stop, cancellation, or an exception escaping retry). The concurrent-start guard in DatafeedLoopV1/V2.start() is unchanged.
Tests — strategy_test.py (ClientPayloadError is classified transient) and abstract_datafeed_loop_test.py (_running reset after a crash allows restart).
Verification
pytest tests/core/retry/strategy_test.py tests/core/service/datafeed/abstract_datafeed_loop_test.py → 43 passed (incl. 2 new).

Compatibility
Both changes are small and backward-compatible. No public API changes.

PR #402, opened 2026-09-01
SEARCH-3102 add semantic search to bdk (#374)

broHeryk:
### Description
Closes SEARCH-3102

We've got ability to use semantic search instead alongside exact match search. This PR adds a method to handle it.

### Dependencies
None

### Checklist
- [x] Referenced an issue in the PR title or description
- [x] Updated the documentation in [docsrc folder](../tree/main/docsrc)


sebastientosello:
Lets take a different path to get 3.x aligned with main branch and ship the semantic search bit. Here is what we can do ;
- Lets replace this PR against 3.x branch with only the semantic search related work
- I will create branch and tag of current main as 2.x
- Then lets open a PR from 3.x to main and push a beta release for internal testing/best effort validation

sebastientosello (approved):

